PR #5, opened 2015-11-02
Fixing up FciGes a bit more for jmo.

jdramsey:
Fixing up FciGes a bit more for jmo.


jdramsey:
Bug fixes for FCIGES.

PR #79, opened 2015-12-03
development

jdramsey:
Code cleanup for sem package.


jdramsey:
All of the tests pass, I checked. These are the same sorts of changes I made to the other packages--javadoc fixes, IntelliJ analyze fixes, a couple of algorithms bug fixes, one fairly major in the SemIm class (the wrong thing was being returned). I'm sure more can be done; I just wanted to make a pass through.

jdramsey:
Oh I fixed the bug in the large scale simulator that Kevin B. pointed out.

jdramsey:
Right--there was also a bug in the SemIm method returning values of parameters where it was returning regression estimates. It may have seemed like a good idea at the time, but it was, in fact, evil. I commented it out.

jdramsey:
I can't tell--were my changes to the graph package just included in my pull request? There's a note above in the GibHub interface.

jdramsey:
Thanks!

PR #94, opened 2015-12-08
code-cleanup

jdramsey:
Couple of bug fixes, plus refactoring the random graph methods.


jdramsey:
Kevin, this has in it the refactored random graph methods. I think these are in good shape if you agree to the method signatures.

jdramsey:
Kevin--I adjusted your code to the cleaned up random graph methods. Quick is not longer there; that was not meant to be public. In any case, the random forward edge method is quick.

PR #156, opened 2016-03-14
Cli validations

kvb2univpitt:
Added the following data validation to FGS cli:
1) Missing and unparseable values.
2) Non-unique variable names.
3) Non-unique variables
4) Zero variance.


jdramsey:
This looks great! I'll assume the tests still pass but run them anyway in a bit.

PR #188, opened 2016-05-13
autistics

jdramsey:


jdramsey:
For Dan. Also, turned off some tests that aren't working.

PR #239, opened 2016-07-21
algcomparison

jdramsey:
Some cleanup of algorithm tool.


jdramsey:
Some further cleanup of algorithm tool responding to Dan's remarks.

PR #263, opened 2016-07-29
dan

dmalinsk:
Adding time series algorithms and related methods.


dmalinsk:
If you guys see any problems with this, email Dan or me. Joe

PR #294, opened 2016-08-24
experimental

jdramsey:
Large commit of GUI work.


jdramsey:
Notes. 

As for functionality changes, there are a few; complain if they don't make sense. Among these:

* I put the templates in the File menu and stopped calling them "templates". I hope that will make it easy for people to find them.

* I fixed Knowledge throughout (I think). At least all of the issues that had been reported.

* I made a new box for Simulation (in the Data box) where you can just select the type of simulation you want and set its parameters. I think this will be much more intuitive. You can still do it the old way (in fact you need to sometimes), but now it's a one stop shop. The Simulation box makes a simulation a la the algorithms comparison API, so they're synced.

I also made a new box called General Algorithm, which is a single algorithm box you can attach to loaded data or a simulation and select algorithms using drop-down menus. Much easier. You can change you mind about which algorithm to use without making a new box. Not all of the algorithms are in there yet; that's going to take some work. All of the algorithms in this box use the algorithm comparison API, so that's all settled now. The algorithms that aren't in there yet haven't been ported yet. But for the ones that are, it's all synced.

The simulation box can simulation multiple data set, possibly with different graphs for each, and the new search box can search over multiple data sets, producing multiple graphs, so the old right click simulation menu items are not longer needed. So I took them out. If you add a comparison box with simulation and algorithm as input, it will run through data sets and graphs and produce output. So all fo the comparison gadgets now can handle multiple models as input and will produce an appropriate output.

Doing it this way has the added benefit that simulation styles could be included that couldn't be included before. You may see a few.
 
I also changed the order of the buttons to put data/simulation and Search at the top, since I observed this is what people naturally wanted to do. All of the other stuff is down below now. You can still do everything from before though.

Like I told Kevin, I moved Fgs2 to Fgs and Fgs to FgsOld. We're trying ot publish a paper with Fgs (the old Fgs2).

There were other issues I fixed as well, which you can find in the issue tracker.

Hopefully, that's a little better. I know still not ideal, but if like I said if you find problems make issues and next time I'm manic I'll fix them. :)

jdramsey:
Also, I added back the old JavaHelp, which has old Help files displayed; these will have to be updated. But we needed hooks from the Search box for the little circle-I icons, and this is where the text should naturally be that they should be linked to.

PR #322, opened 2016-09-29
Tetrad enable verbosity

kvb2univpitt:
Enable the new FGS and GFCI interfaces to write messages to file.


jdramsey:
This looks fine.

PR #353, opened 2016-11-28
experimental

jdramsey:
PAG coloring code.


jdramsey:
This pushes several bugs fixes but mainly the PAG graph coloring function, in GraphUtils. There is a test for it in TestGraphUtils and it is used in the interface, in AbstractWorkbench.

jdramsey:
I was trying to cherry pick the stuff to commit but ended up not including everything needed for the PAG coloring, so I'm pushed all of my work, including several bug fixes to various pieces of code, plus some recent work on mixed variables and incremental improvements to several algorithms. Fixes to graph error counting by Ruben are included. For work on algorithms, see commits. For PAG coloring, in edu.cmu.tetrad.graph.GraphUtils there is a method called addPagColoring which adds dashed strokes and/or green coloring to edges using the semantics Peter provided. In TestGraphUtils there is a test, testPagColoring, which tests these colorings against the true graph (with latents) using a test devised by Peter. In the GUI, if you make data and run a PAG algorithm on it, the PAG will be colored. To make this happen, PAG algorithms need to "stamp" their output graphs as PAGs--there is a setPag(true) method for that, which is called for all PAG algorithms. The PAG coloring in the GUI is done in the AbstractWorkbench class, if the graph if isPag() returns true:

            if (graph.isPag()) {
                GraphUtils.addPagColoring(new EdgeListGraph(graph));
            }

harryhoch:
looks good.

PR #369, opened 2016-12-16
change output target of causal-cmd

espinoj:
@harryhoch try this and if it works please merge


espinoj:
Cancelling this until we separate from tetrad code base

PR #478, opened 2017-06-16
CG structure prior fix

bja43:
added penalty discount back, multiplied SP by 2, and fixed p in SP @jdramsey


jdramsey:
For some reason there's a test failure with Travis. I pulled the changes over to 'joebugfixes' and ran the test. No failure. Pushed to 'development'. We'll see what Travis comes up with.

jdramsey:
This is included in the 'joebugfixes'.

Issue #486, opened 2017-06-19
Restore Mimbuild in the interface.


jdramsey:
In the config file, this bit of code needs to be uncommented:

            <model name="MIMBuild" acronym="MIMBuild" help="mimbuild"
                   category="Multiple Indicator Model searches">
                <logger>
                    <event id="info" description="Information" default="on"/>
                    <event id="details" description="Misc Fine Details" default="on"/>
                    <event id="maxmodel" description="Max P Value Model" default="on"/>
                </logger>
                <model-class>edu.cmu.tetradapp.model.MimBuildRunner
                </model-class>
                <editor-class>edu.cmu.tetradapp.editor.MimbuildEditor
                </editor-class>
            </model>

jdramsey:
@yuanzhou 
@espinoj 

Maybe there's a better interface for Mimbuild, but this one at least works. Here's a screenshot.

<img width="785" alt="screen shot 2017-06-19 at 6 05 03 pm" src="https://user-images.githubusercontent.com/9853255/27308347-fe096368-551b-11e7-8589-b9078220aabd.png">

jdramsey:
Also, @ekummerfeld @ps7z

yuanzhou:
@jdramsey Do I just uncomment the config section and test things out or is there anything specific you want me to look at?

jdramsey:
@yuanzhou I think just uncomment it and try it out, as you say. It should just work; it was working before, and it worked for me just now. I thought it better if you did the honors though, so you weren't surprised.

yuanzhou:
@jdramsey no problem. Since I'm also fixing another Tetrad bug, I may combine this issue and the one I'm working on in the same pull request later.

jdramsey:
@yuanzhou Great, thanks!

yuanzhou:
I was able to get the Mimbuild work. Also noticed the weird highlighting issue that we noticed int he knowledge tier box.

![mim](https://user-images.githubusercontent.com/195873/27353476-25af1ec4-55d2-11e7-8e98-56518a262eab.PNG)
![mm](https://user-images.githubusercontent.com/195873/27353478-2683b4c2-55d2-11e7-9d48-e38909eaf6bb.PNG)

@jdramsey for some reason, I could only get it work once. I tried several times later with multiple searches but just couldn't reproduce. What is the proper procedure to use the Mimbuild?

yuanzhou:
[...]

Issue #477, opened 2017-06-15
Unable to Delete Selected Column From Data Box


kvb2univpitt:
User unable to delete selected column via "Delete Selected Rows or Columns" from the "Edit" menu.

jdramsey:
This used to work. I wonder how it got broken?

yuanzhou:
I tested this and couldn't "Delete Selected Rows or Columns".  I saw some errors in the console. I'll work on the fix.

yuanzhou:
@jdramsey the broken code was calling `removeCols()` at https://github.com/cmu-phil/tetrad/blob/data-box-fix/tetrad-lib/src/main/java/edu/cmu/tetrad/data/BoxDataSet.java#L1109 and there's index out of range error which I couldn't figure out.

I fixed the bug by switching to `removeColumn()` to remove each selected individual column. https://github.com/cmu-phil/tetrad/blob/data-box-fix/tetrad-lib/src/main/java/edu/cmu/tetrad/data/BoxDataSet.java#L596

I've commented the old code here: https://github.com/cmu-phil/tetrad/blob/data-box-fix/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/TabularDataJTable.java#L296

But I don't see any issue with `removeCols()`  during debugging. Can you also take a look just so we are on the safe side. Thanks.

jdramsey:
@yuanzhou That's a perfectly good fix. I'll look at the removeCols method, though now that you've fixed it I'll have to think of a newly clever way to break it. :)

PR #488, opened 2017-06-21
Data box fix

yuanzhou:
This PR addressed the issue #477 and #486 

@jdramsey please review. It can be auto merged by clicking the green button on this page. Thanks!


jdramsey:
@yuanzhou  This looks good to me. Thanks for fixing both of those things. The only thing I can think of is that maybe the minimum number of clusters should be 2--someone may simply want to know if there's an edge between two clusters or not.

jdramsey:
@yuanzhou I'll merge this. The cluster minimum can be discussed later. I'll ask Peter and Erich.

yuanzhou:
@jdramsey thanks, the cluster minimum can be changed at any time.

PR #496, opened 2017-07-08
Algcomparison with r

jdramsey:
In this pull request, I do the following.

1. I create a large simulation study for the linear, Gaussian case. See

https://github.com/cmu-phil/comparison-data

Conditions 1. I include several algorithms in Tetrad as well as from the pcalg and bnlearn packages in R and the BNT package in Matab. To this end, I modify the Comparison class in algcomparison. I create objects of the type ExternalAlgorithm that are able to read in graphs and elapsed time files from various platorms/algorithm and include them in the comparison table. I make a separate method in Comparison to load these in and run them. See the file Comparison1.java in package edu.cmu.tetrad.test.joes.Condition1.

2. I combine all fo the PC-style algorithms into one meta-PC algorithm with switches. This is called PcAll. I adjust all of the algorithm comparison PC-algorithm wrappers so that they use this, and I also provide an algorithm comparison wrapper for the meta-algorithm.


jdramsey:
I'll go ahead and merge this.

PR #502, opened 2017-07-25
algcomparison_with_r

jdramsey:
Sundry recent work from Joe.


jdramsey:
Some of this was a previous canceled pull request I did:

In this pull request, I do the following.

I create a large simulation study for the linear, Gaussian case. See
https://github.com/cmu-phil/comparison-data

Conditions 1. I include several algorithms in Tetrad as well as from the pcalg and bnlearn packages in R and the BNT package in Matab. To this end, I modify the Comparison class in algcomparison. I create objects of the type ExternalAlgorithm that are able to read in graphs and elapsed time files from various platorms/algorithm and include them in the comparison table. I make a separate method in Comparison to load these in and run them. See the file Comparison1.java in package edu.cmu.tetrad.test.joes.Condition1.

I combine all fo the PC-style algorithms into one meta-PC algorithm with switches. This is called PcAll. I adjust all of the algorithm comparison PC-algorithm wrappers so that they use this, and I also provide an algorithm comparison wrapper for the meta-algorithm.

I that's the only relevant stuff for you guys.

yuanzhou:
Thanks Joe! I'll test the combined PC algorithms. And maybe @kvb2univpitt and @chirayukong can help check out other pieces of your changes. Normally @espinoj is the merge master, but now he's in Africa these weeks.

yuanzhou:
@kvb2univpitt @chirayukong below are what @jdramsey replied regarding this PR for review/test purposes. He didn't hit "reply-all" so it only went to me. I'm posting them here so we know what to look at before merging.
>I guess the main thing is to make sure all of the interface stuff still works.
I did make one change to FGES that I should note, though it is innocuous: I moved the method for calculating unconditional d-separations int the FGES class itself.
Other than that I think it's good. I didn't specifically change any of the other algorithm code. There may have been minor tweaks here and there.
Oh, I change the Comparison code in a few places, but it all still runs. Also, I added PcAll.
The psychometrics library I had added I took back out, so don't worry about that.
The StARS and stability code are still under development, but they don't affect anything else.
The last several commits were for the paper with Bryan and Greg and me.

I can merge this PR later.

jdramsey:
Sorry for not hitting reply all. I'm bad about that. I will be by tomorrow like I said if I can help at all, after the 9-11 meeting.

yuanzhou:
@jdramsey I merged the `algcomparison_with_r` into `development` locally for testing, and I noticed couple of bugs:

1. Load some data, then do a search, using all default search settings (PC with default parameters), then hit the search button. This doesn't do any search and in the console, I could see some errors:
capture4

2. Add a knowledge box to the data model, the default selection is "Measurement Model" and if I just hit the OK button, I saw a popup with empty message. When I close that popup, I could some errors.

capture
capture2

jdramsey:
@yuanzhou I'll sit at Cafe Mocha after the meeting and try to fix these problems.

Joe

jdramsey:
@yuanzhou OK I fixed those two problems, pushed the changes. Any other problems?

Joe

yuanzhou:
Thanks @jdramsey for the quick fix. Another error I noticed in console is when I added a new graph box, link it from a search result. Double clicked and chose "Graph" from the list, I could see the graph but the console throws this error:

capture

The search was for a continuous dataset with PC and its default settings.

jdramsey:
@yuanzhou That wasn't caused by my commits, but you can fix it by adding a null pointer check as follows, AbstractWorkbench, line 2660:

		public final void componentMoved(ComponentEvent e) {
			Component source = (Component) e.getSource();
			Rectangle bounds = source.getBounds();

			if (source instanceof DisplayNode) {
				Node modelNode = (Node) (workbench.getDisplayToModel().get(source));

                               **_if (modelNode == null) return;_**

				int centerX = bounds.x + bounds.width / 2;
				int centerY = bounds.y + bounds.height / 2;

				modelNode.setCenterX(centerX);
				modelNode.setCenterY(centerY);

				workbench.adjustPreferredSize();

				// This causes wierdness when nodes are dragged off to the
				// right. Replacing with a scroll to rect on mouseup.
				// jdramsey 4/29/2005
				// workbench.scrollRectToVisible(bounds);
			}
		}

yuanzhou:
@jdramsey  Just fixed and tested, that exception message is gone. I've also pushed it to your ` algcomparison_with_r` branch.

I'm going to merge this PR shortly and if we find other bugs/issues later, we'll do another fix.

yuanzhou:
@jdramsey I'll leave it to you to delete your `algcomparison_with_r` branch just in case you are using it for other purposes. In general a merged branch should be deleted right after it's merged.

jdramsey:
@ok, thanks!

PR #517, opened 2017-09-19
Causal cmd enhancements

kvb2univpitt:


yuanzhou:
Checking it now, thanks!

PR #519, opened 2017-09-20
Annotation additions

kvb2univpitt:
Zhou, I applied the new annotation classes to the tetrad-gui already.  It is working now.  Please test it.


yuanzhou:
@kvb2univpitt awesome! Testing now. Thanks!

yuanzhou:
@kvb2univpitt I noticed that `TetradAlgorithmAnnotations.getInstance().getNameWrappers()` doesn't return an ordered list. I tested several times and the ordering of the algos kept changing. We need to use TreeMap to store the store the `nameWrappers`. Same for `acceptKnowledgeNameWrappers` and `acceptMultiDatasetNameWrappers`.

And apply the same change to `TetradScoreAnnotations` and `TetradTestOfIndependenceAnnotations`.

yuanzhou:
Merge now and Kevin will order the list in a new branch.

PR #524, opened 2017-09-26
Annotation refactoring

kvb2univpitt:
@yuanzhou You should now be able to add multiple values to attribute datatype  for independence test and score like this:

```java
dataType = {DataType.Continuous, DataType.Covariance}
```
Please use the list @jdramsey sent yesterday to add covariance and graph data type to scores and independence tests.


yuanzhou:
Tested and merged

PR #540, opened 2017-10-23
Refactor bootstrapping

chirayukong:


kvb2univpitt:
@chirayukong The description says the minimum value is 1, but it's actually 0.  I'll make that change and merge.

Issue #548, opened 2017-10-28
Data parsing issue.


jdramsey:
If you try to load this dataset in the GUI by selecting Discrete and quote character = ", it hangs and never comes back.

[sachs.discretised.txt](https://github.com/cmu-phil/tetrad/files/1424548/sachs.discretised.txt)

chirayukong:
<img width="948" alt="screen shot 2017-11-01 at 5 08 15 pm" src="https://user-images.githubusercontent.com/4652250/32298067-9056c42a-bf27-11e7-9e5d-bd597f628884.png">
<img width="948" alt="screen shot 2017-11-01 at 5 08 28 pm" src="https://user-images.githubusercontent.com/4652250/32298077-99bffdd8-bf27-11e7-9bdd-f841f91acab0.png">
<img width="947" alt="screen shot 2017-11-01 at 5 11 11 pm" src="https://user-images.githubusercontent.com/4652250/32298108-b07ce73e-bf27-11e7-9c0f-9dfea6ea5567.png">
It seems once you loaded the dataset, Tetrad froze. The error is in the data reader:
Exception in thread "Thread-8" java.lang.ArrayIndexOutOfBoundsException: 0
	at edu.pitt.dbmi.data.reader.tabular.AbstractDiscreteTabularDataFileReader.extractVariables(AbstractDiscreteTabularDataFileReader.java:578)
	at edu.pitt.dbmi.data.reader.tabular.VerticalDiscreteTabularDataReader.readInDataFromFile(VerticalDiscreteTabularDataReader.java:41)
	at edu.pitt.dbmi.data.reader.tabular.AbstractTabularDataFileReader.readInData(AbstractTabularDataFileReader.java:46)
	at edu.pitt.dbmi.data.reader.tabular.AbstractTabularDataFileReader.readInData(AbstractTabularDataFileReader.java:56)
	at edu.cmu.tetradapp.editor.DataLoaderSettings.loadDataWithSettings(DataLoaderSettings.java:1036)
	at edu.cmu.tetradapp.editor.LoadDataDialog.loadAllFiles(LoadDataDialog.java:846)
	at edu.cmu.tetradapp.editor.LoadDataDialog.access$28(LoadDataDialog.java:843)
	at edu.cmu.tetradapp.editor.LoadDataDialog$7$1.run(LoadDataDialog.java:616)
	at java.lang.Thread.run(Thread.java:745)
@kvb2univpitt

[...]

PR #563, opened 2017-11-09
Updated data-reader version to the latest.

kvb2univpitt:
Updated the data reader dependency to the latest to fix issue #548.


jdramsey:
Great, thanks!

PR #610, opened 2017-11-15
Fix issue 543

chirayukong:


espinoj:
Kevin will work on this more

PR #613, opened 2017-11-28
Fix issue 597

chirayukong:
# Conflicts:
#
tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/SimulationEditor.java

PR #619, opened 2017-12-06
Fix issue 543 kevin

kvb2univpitt:
Latest GUI and annotation utilities fixes.


espinoj:
Does this supercede pull #613 ?

kvb2univpitt:
Not sure.  This is a question for @chirayukong

PR #627, opened 2017-12-20
FASKGFCI

jdramsey:


jdramsey:
Sorry, I just noticed there are some issues. Resolving.

I am adding the dependency Kong pointed out that prevents Tetrad GUI from being launched under 1.9. This:


<dependency>
    <groupId>javax.xml.bind</groupId>
    <artifactId>jaxb-api</artifactId>
    <version>2.3.0</version>
</dependency>

I am fixing the conflicts noted above.

Fixed a couple of unit tests.

Now everything seems to be HAPPY.

jdramsey:
@kvb2univpitt @jeremy

(Mostly Kevin.)

I edited the branch https://github.com/cmu-phil/tetrad/tree/v6.4.0-task-timeout to check interrupted status for all fo the algorithms being displayed in the Tetrad GUI, with one exception: PCStableMax, which was included in PC All. I commented out the annotation for that one.

Should I continue to check interrupted status for the algorithms not displayed in the Tetrad GUI?

Most of these were easy, but a couple of them, like BPC, I had to check the hell out of because they are so slow and prone to hanging. :(

In ChoiceGenerator, which iterates through all the choice of a choose b, I added a check to stop the iteration if the thread is interrupted. So all uses of choice generator throughout the program will behave correctly. This is important because this is the main source of programs hanging, so even algorithms I don't explicitly check interrupted status should be pretty OK. But I added explicit checks to a lot of the choice generator while loops for clarity. Should these explicit checks be removed? I'm assuming checking a boolean is really fast.

I wish I could do this with for loops. In fact, I wish Java had just done that with all for loops! But alas.

Joe

Issue #633, opened 2018-01-07
Simulation box does not use input graph


mglymour:
In 6.4.0-0, downloaded today, the simulation box will not simulate data based on an input graph. Simulation boxes with graph input boxes act as though they are parentless simulation boxes.
<img width="1430" alt="screen shot 2018-01-07 at 2 04 24 pm" src="https://user-images.githubusercontent.com/15677053/34653042-30536350-f3b4-11e7-970b-1b3afd2a3159.png">

kvb2univpitt:
Fixed.

PR #634, opened 2018-01-08
Fixed bug where simulation does not use input graph.

kvb2univpitt:
Fix for issue #633


jdramsey:
It seems right. That used to work; not sure what went wrong. But this seems like a good workaround.

kvb2univpitt:
@jdramsey It's actually not a workaround. I mistakenly added an extra line of code that was never there in the old code when I was working on the GUI part of it.  That was my mistake.

jdramsey:
@kvb2univpitt Oh OK. Thanks!

Issue #658, opened 2018-01-12
Estimating SEM from correlation matrix


scheines:
The estimator refused to do its thing when the data is a correlation matrix for a SEM PM.  Below the data file - but I can't attach a .tet file.  Graph is Pre--> Final, Pre --> Print --> Vol --> Quiz --> Final
[pitt_newer.txt](https://github.com/cmu-phil/tetrad/files/1626321/pitt_newer.txt)

espinoj:
In development branch covariance file loads ok.  I recall that we did some bug fixes in covariance loading on this round.

espinoj:
Discussed with @jdramsey.  I'm making a ScoreType enum class that will be used everywhere.

espinoj:
@scheines today's development build has the fix
https://cloud.ccd.pitt.edu/nexus/content/repositories/snapshots/edu/cmu/tetrad-gui/6.4.0-SNAPSHOT/tetrad-gui-6.4.0-20180116.180138-162-launch.jar

PR #660, opened 2018-01-12
Replaced SemIm.ScoreType and ComparisonParameters.ScoreType with a ne…

espinoj:
…w enum class ScoreType.  Fixes estimator can't estimate bug #658


espinoj:
Unfortunately because of class changes can't load old tet files again.  But please check that estimator can estimate - works for me.  Also since this effects comparison should do a run of the tool to make sure it didn't break that - @bja43 or @jdramsey

espinoj:
Gonna merge so we can get this to Richard

Issue #650, opened 2018-01-11
Clark: R1,R2, and R3 do not work.  Remove R4


espinoj:
R1 does not work.

R2 does not work, R3 does not work

R4 should be removed

yuanzhou:
Do we want to remove the R4 class file or just the annotation but still keep the class?

espinoj:
for R4 and EB keep the class.  Check with @kvb2univpitt about the experimental annotation which we will use later as an option to show all available algorithms.

yuanzhou:
Fixed via pull request #667

Issue #645, opened 2018-01-11
Clark: EB should be removed


espinoj:

yuanzhou:
Link this to https://github.com/cmu-phil/tetrad/issues/650

yuanzhou:
Addressed via pull request #667

PR #667, opened 2018-01-17
645 and 650 fixes zhou

yuanzhou:
This PR addressed issues of #645 and #650. EB and R4 are hided from the Algo chooser and R1, R2, R3 and all other pairwise algorithms are working.

@jdramsey @kvb2univpitt please test and merge if everything looks good. Thanks!


yuanzhou:
Let me fix the test error first.

yuanzhou:
Turned out that `TetradSerializableUtils.serializeClass()` forces the "serialVersionUID" to be 23L in each algorithm class, and I generated the UID instead of using default one, which failed the test.

kvb2univpitt:
Looks ok.

Issue #592, opened 2017-11-14
Clark - RSkew and RSkewE


espinoj:
RSKew and RskewE

Rskew should be removed. Only RskewE is empirically applicable. Description needs to be altered to explain why.

yuanzhou:
Same as what I did with EB (#645 ) and R4 (#650 ), I'll just add the `@Experimental` annotation to hide it from the algo chooser and leave the class file there.

jdramsey:
Wait, RSkew works! It's really good! The best of the lot! RSkewE is worse! I can explain!

Joe

@espinoj @yuanzhou

yuanzhou:
RShew is back now https://github.com/cmu-phil/tetrad/commit/89a6a301a0fcee36fe9da021d2bb7d810f9f6f0d

Currently we have R1, R2, R3, RSkew, RShewE, Skew, SkewE under the pairwise category.

jdramsey:
That sounds OK. Like I said, I can send a reference if you like; we wrote all of this up with simulation results and published it in NeuroImage.

espinoj:
Closing this.  Clark erroneously said to remove RSkew

PR #676, opened 2018-01-19
Hide RSkew to address #592

yuanzhou:
#592 fix. I'll just merge.


jdramsey:
Please don't hide RSkew!

@espinoj @yuanzhou

Issue #497, opened 2017-07-10
Cut/copy/paste of cells in data box broken


mglymour:
When I attempt to copy and paste a column of cells from one data box to another, I sometimes get an off-by-one error (the wrong column is copied), and sometimes fail to paste altogether. In some cases, the correct column will be cut, but then the column next to it is what pastes, losing the data in that column entirely.

I'm using a version downloaded from development on 07/10/2017.

jdramsey:
Add this fix to https://github.com/cmu-phil/tetrad/pull/703.

PR #703, opened 2018-01-29
time.lag.fixes2

jdramsey:
Several more time series fixes.


jdramsey:
I was working on this, so I self-assigned it. I have fixed the mislabeling problem by removing the variable sorting operation from the time lag filter. (That was a bad idea anyway.) The Time Lag with Index I replaced with just an Add Index manipulation. It's usually not reasonable for someone to add an index to the data and then also lag it. (Why would you do that? The Time variable will be disconnected anyway.) But if someone _wants_ to do that they can add another box to lag the data with the index variable in it.

I fixed the problem where if you put TimeLag->Knowledge double click knowledge, showing the time lag knowledge, then destroy and recreate the knowledge, it wouldn't give you back the knowledge. Still, if you destroy and recreate the time lag data, then double-click knowledge, there are no tiers. But if you then destroy and recreate the knowledge the tiers appear again. Not sure what's happening.

@jeremy

jdramsey:
Added a fix for https://github.com/cmu-phil/tetrad/issues/497.

Issue #705, opened 2018-01-29
Fix capitalization of these algorithms...


jdramsey:
In the Search box.

Bpc -> BPC
FgesMb -> FGES-MB
Fofc -> FOFC
Ftfc -> FTFC
TsImage -> TsIMaGES

@cg09 @espinoj

yuanzhou:
I'll work on this renaming and sync with @kvb2univpitt  on the API and Causal CMD.

yuanzhou:
Fixed via pull request #708

PR #708, opened 2018-01-31
Algo names captalization, addressed #705

yuanzhou:


yuanzhou:
@kvb2univpitt I'll rebuild the REST API to use the updated Tetrad later. I guess you'll need to do the same thing with Causal CMD.

kvb2univpitt:
Looks good.

Issue #709, opened 2018-01-31
Error message with TsIMaGES


yuanzhou:
When I ran TsIMaGES on a single continuous or discrete dataset, I got this error message:

 
![capture1](https://user-images.githubusercontent.com/195873/35644674-96cae568-0697-11e8-8bdd-d00718b7d0a7.PNG)

![capture2](https://user-images.githubusercontent.com/195873/35644677-98c2ae64-0697-11e8-9205-e137ac43dbd2.PNG)

yuanzhou:
Any suggestions @jdramsey ?

jdramsey:
@yuanzhou Well here's a mistake; I wonder if this would fix it? This code block:

        if (score instanceof SemBicScore) {
            SemBicScoreImages gesScore = new SemBicScoreImages(dataModels);
            gesScore.setPenaltyDiscount(parameters.getDouble("penaltyDiscount"));
            IndependenceTest test = new IndTestScore(gesScore);
            search = new TsGFci(test, gesScore);
        } else if (score instanceof BDeScore) {
            double samplePrior = parameters.getDouble("samplePrior", 1);
            double structurePrior = parameters.getDouble("structurePrior", 1);
            BdeuScoreImages score = new BdeuScoreImages(dataModels);
            score.setSamplePrior(samplePrior);
            score.setStructurePrior(structurePrior);
            IndependenceTest test = new IndTestScore(score);
            search = new TsGFci(test, score);
        } else {
            throw new IllegalStateException("Sorry, data must either be all continuous or all discrete.");
        }

When it says  if (score instanceof BDeScore), that should be BdeuScore. You can see down further Bdeu is meant.

yuanzhou:
Thanks @jdramsey, you have eagle eyes. I'll test it out.

kvb2univpitt:
@yuanzhou  @jdramsey This issue is fixed.  However, TsIMaGES is still not able to run because there's a bug in edu.cmu.tetrad.search.TsFges2.  I will create a different issue for this bug.

jdramsey:
Thanks!

yuanzhou:
I noticed some bugs during testing, posted in the pull request thread #721 

Also link this issue to #722

kvb2univpitt:
@yuanzhou That bug is not related to this issue.  That bug is has to do with algorithm handing invalid indices.

yuanzhou:
This issue has been addressed via pull request #721

PR #721, opened 2018-02-07
V6.4.0 fix tsimages

kvb2univpitt:
Fix for issue #709

### DIFF
--- a/tetrad-gui/src/main/java/edu/cmu/tetradapp/model/GeneralAlgorithmRunner.java
+++ b/tetrad-gui/src/main/java/edu/cmu/tetradapp/model/GeneralAlgorithmRunner.java
@@ -53,6 +53,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * Stores an algorithms in the format of the algorithm comparison API.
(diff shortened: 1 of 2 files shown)


yuanzhou:
Just tested, and when I chose a different score other than "SEM BIC" or "BDeu", I was supposed to see the error message in 

````
public TsImages(ScoreWrapper score) {
        if (!(score instanceof SemBicScore || score instanceof BdeuScore)) {
            throw new IllegalArgumentException("Only SEM BIC score or BDeu score can be used with this, sorry.");
        }

        this.score = score;
    }
````

But I didn't. And I still saw this popup:

capture

**Comment by kvb2univpitt** [comment]
@yuanzhou The reason why you still see that error is because the logic only accepts either SEM BIC score or BDeu score.  See code below.  Any other score will give you that error message.  What's odd is that BDeu score does not show up in the list of scores.  And, other scores are showing up.

```java
if (score instanceof SemBicScore) {
    ...
} else if (score instanceof BdeuScore) {
    ...
} else {
    throw new IllegalStateException("Sorry, data must either be all continuous or all discrete.");
}
```

@jdramsey Joe, what scores are valid for this algorithm?

kvb2univpitt:
Ahh....I remember now.  Scores are based on data type.  Thus, we see other scores in the list.  Should TsIMaGES take other scores?

yuanzhou:
@jdramsey I think I agree with @kvb2univpitt that this fix should be merged since the purpose of this pull request is to fix the error message. We'll need to fix the underlying bug separately.

jdramsey:
@yuanzhou I have an idea. Why not just change the text? Instead of this:

if (score instanceof SemBicScore) {
    ...
} else if (score instanceof BdeuScore) {
    ...
} else {
    throw new IllegalStateException("Sorry, data must either be all continuous or all discrete.");
}

say this:

if (score instanceof SemBicScore) {
    ...
} else if (score instanceof BdeuScore) {
    ...
} else {
    throw new IllegalStateException("Sorry, the score must be SemBicScore (for the continuous case) or BDeu score (for the discrete case).");
}

kvb2univpitt:
@jdramsey @yuanzhou I think the preferred way would be filter it out in the list of scores.  I'll make this as a separate issue.

clip

Issue #719, opened 2018-02-06
Can't change the value of alpha in the Regression box


chirayukong:
The default value is 0.001 but you can't change to any other value. Once you clicked the "execute" button, the button wouldn't work again. You need to destroy the model in order to make some change in the UI.
<img width="1160" alt="screen shot 2018-02-06 at 2 54 03 pm" src="https://user-images.githubusercontent.com/4652250/35881214-1da60698-0b4e-11e8-9de3-c6fbd03e4fff.png">

chirayukong:
I tried to change the alpha from 0.001 to 0.0001 but nothing happened. The attachment is the tetrad session's file. @jdramsey 

<img width="932" alt="screen shot 2018-02-07 at 3 37 02 pm" src="https://user-images.githubusercontent.com/4652250/35940403-f8055324-0c1c-11e8-9847-752061d3c984.png">
[multiple_linear_regression.tet.zip](https://github.com/cmu-phil/tetrad/files/1704658/multiple_linear_regression.tet.zip)

kvb2univpitt:
Fixed.
@chirayukong  You can try it out now.

chirayukong:
It works. Thanks!

PR #725, opened 2018-02-09
V6.4.0 fix alpha regression box

kvb2univpitt:
Fix for issue #719.


jdramsey:
@kvb2univpitt @chirayukong What was the problem?

chirayukong:
There is no setter for alpha. So, you can't change the alpha's value. @kvb2univpitt added it and it works.

jdramsey:
Nice. Thanks!

Issue #731, opened 2018-02-12
Labels of Search Output


cg09:
I think this is a repeat request:  the output of a search should contain the name of the algorithm used in the search.

yuanzhou:
This turned out to be not just a simple label change. I'll discuss with @kvb2univpitt to come up with a good plan.

kvb2univpitt:
Done.  I also added algorithm in the parameter panel.

![algo](https://user-images.githubusercontent.com/10729309/36175075-48a98048-10dc-11e8-8158-4b18c86d82ee.png)
![param](https://user-images.githubusercontent.com/10729309/36175076-48b89646-10dc-11e8-9a09-8212f0418222.png)

cg09:
great! many thanks.

jdramsey:
This looks great!  Joe

PR #735, opened 2018-02-13
V6.4.0 add algo name

kvb2univpitt:
Fix issue #731.


yuanzhou:
Tested and looked great! Thanks!

PR #745, opened 2018-02-21
Fixed parameter list to reflect simulation drop-down list.

kvb2univpitt:
Misc. code clean up.


jdramsey:
I fixed one small thing--time series simulations should not be marked as 'fixed'--I think was accidentally left over (I think by me) from the Boolean Glass case, which was eliminated. I think it's good now though.

jdramsey:
@kvb2univpitt

PR #746, opened 2018-02-21
Commented out all of the Meek Orientations menu items.

jdramsey:


jdramsey:
This just comments out some lines in the code Clark and I thought were a little confusing to the reader. These were the menu items that applied the Meek orientation rules to a given graph. Usually, no effect could be seen, and to make a graph where you'd see the effect you'd need to really work at it, so...confusing. It's just an internal operation.

Issue #737, opened 2018-02-14
More on the simulation box


cg09:
clicking the simulation box automatically generates a sample of 1,000. Then the user has to figure out that clicking "simulation set up" inside the simulation box allows the user to generate a different sample with a different sample size.   This is illogical.  Clicking the simulation box SHOULD bring up the "simulation set up" window, which then allows simulated data as specified.

jdramsey:
I think this is as simple as just not doing the initial simulation. I think I did that so the data tab would be populated. I think you just need to all the data tab to be unpopulated until a simulation is done.

cg09:
???

jdramsey:
I have a suggestion for this, pull request https://github.com/cmu-phil/tetrad/pull/747. See what you guys think. I simply commented out the lines in Simulation that automatically created simulation data without the user having to click "Simulate." I thought it would break stuff but it did not. Now when you double-click the Simulation box and click on the Data tab or the Graph tab, you find no graph or data there. To get graphs or data sets, you have to click Simulate. I think this is in the spirit of what @cg09 had suggested. I've tried it a bunch of different ways now, and nothing is broken.

But it's just a suggestion--please don't merge this unless it's what you guys would have done anyway, all things considered.

@cg09 @espinoj @kvb2univpitt

espinoj:
Clicking to generate is in line with other box behavior.  Merging.

PR #747, opened 2018-02-21
Commented out the lines that made simulated data before the user clicked 'simulate'.

jdramsey:


jdramsey:
See https://github.com/cmu-phil/tetrad/issues/737. Please don't merge without review--this is just a suggestion.

PR #782, opened 2018-05-21
Forbid n tier

espinoj:
Added feature where in tiers and edges tier will only affect successive tier.  Refactored out Knowledge class.  Moved saveKnowledge to DataWriter class.


cg09:
The basic problem is that prior knowledge constrains a partial order, and
with tiers we can only impose something like linear order constraints.

PR #792, opened 2018-06-01
Recent versions of Java "fix" sorting methods underlying the swing li…

espinoj:
…braries causing general contract warnings.  This reverts to a legacy sorting method that does not generate warnings.


espinoj:
@kvb2univpitt can you please review and merge this so we can release tomorrow?

yuanzhou:
Looks like it's JRE specific since I couldn't reproduce the same error that @espinoj encountered on his Mac. This PR fixes the error so I'll just merge.

Issue #827, opened 2018-06-14
inappropriate version message


harryhoch:
attached session file complains about versioning issues, even when both saved and loaded with 6.5.0.

Java log file and error message screenshot attached.  Rename session file to remove .txt fie. 
[20180613_Cat_CRS_simple_with_diversity.tet.txt](https://github.com/cmu-phil/tetrad/files/2100542/20180613_Cat_CRS_simple_with_diversity.tet.txt)
[log.txt](https://github.com/cmu-phil/tetrad/files/2100543/log.txt)
![armbruster_error_message](https://user-images.githubusercontent.com/174707/41387597-b91fd3cc-6f55-11e8-99ef-ab7ee918ce55.jpeg)

yuanzhou:
@harryhoch when you saved the session, did you use "Save session" or "Save session as" option? It seems the session wasn't saved correctly, that's why we can't reload it. The version message in the popup is not appropriate anyway.

harryhoch:
@yuanzhou, I didn't save it. One of the short course participants did. I have no idea what option was selected. Why does it matter?

yuanzhou:
because from my debugging, I saw a serialization exception and `SaveSessionAsAction.java` was in the stack trace. I guess the session file was not saved correctly.

yuanzhou:
One thing I tested is saving a session from Tetrad 6.5.0 and load it in Tetrad 6.6.0-SNAPSHOT. It worked fine. I did fix the popup message so it doesn't lead the users to try a different Tetrad version. It won't load the corrupted session file regardless the version of Tetrad.

yuanzhou:
I think the root cause of this issue is the corrupted session file instead of Tetrad version. We'll need to create a separate issue when we can reproduce a session file is being saved but he session objects not serialized correctly.

PR #829, opened 2018-06-14
Fix 827 zhou

yuanzhou:
This pull request addressed the inappropriate session loading message reported in #827.


kvb2univpitt:
Need to also apply the fixes to the drag-and-drop functionality.

PR #851, opened 2018-06-28
V6.6.0 obs data factory

kvb2univpitt:
This is the first pass in creating a factory to create interventional datasets.


yuanzhou:
tested and look good. Thanks @kvb2univpitt

PR #878, opened 2018-07-31
Made bootstrapping explicitly shown corresonding nodes -- less confusion

chirayukong:


yuanzhou:
Once this is merged, I'll add those new columns to the bootstrap table rendering.

cg09:
I ask again: how do we get multi-fask into a publicly accessible interface
version of TETRAD? We have uses who need it, we have the code..

Clark

PR #897, opened 2018-08-23
Adjusting the Conditional Correlation code, substituting a revised implementation in for the old one. Adjusting all of the related classes so it works correctly in the interface. Deleting NLO, an ill-fated idea. Adjusted ParamDescriptions to add entries for the parameters Conditional Correlation needs.

jdramsey:
Targeted pull request to update the Conditional Correlation code.


jdramsey:
@espinoj This is just a simply targeted pull request. I branched off of development and moved in the current, updated, code for Conditional Correlation (CCI). I deleted NLO, because it should have been deleted. I adjusted ParamDescriptions for the revised CCI and copied in the classes need to correctly run CCI in the GUI. That's all. We have a user who would like to used this ASAP; could it be published if it's OK? Thanks--Joe

jdramsey:
@espinoj @kvb2univpitt I just noticed I hadn't pushed all of the parameter adjustments in ParamDescriptions. Just did it. Sorry.

jdramsey:
I will send them a jar.

PR #904, opened 2018-09-04
Fixed graph editor layout on full screen

yuanzhou:
This pull request fixed the big empty spaces in graph editor between the left-side toolbar and the right-side graph display. It also made the resize of bootstrap table more flexible.


chirayukong:
Looks good!

yuanzhou:
Thanks @chirayukong

PR #910, opened 2018-09-06
KCI files for development

jdramsey:
Added the (Java) KCI along with related files. Adjusted the parameters for the needed ones for KCI.


jdramsey:
@espinoj These are the files for KCI; it should now work in the interface as an independence test. I set the default for "verbose" to "true". The idea is the user can watch the independence tests being done in the console, or they can open up the logging and see them there, to break the monotony of waiting for the method to come back (i.e. mark its progress).

PR #917, opened 2018-09-14
Generalize sampling and bootstrapping

chirayukong:


jdramsey:
What's the goal here? What should I check?

chirayukong:
> What's the goal here? What should I check?

The goal is to replace bootstrapping parameters with the resampling parameters that allow users to specify a number resampling iterations, a size of resamples, and whether resampling with replacement (bootstrapping) or not. :-)

jdramsey:
Great--let me check.

PR #938, opened 2018-11-19
Adding a print statement for model (BIC) score for FGES. It now prints BIC = ... after the search is done.

jdramsey:


jdramsey:
Jeremy, This is for the request by Jinling to have the BIC score printed when running  FGES. I tested it; it appears to work. (When I say I tested it, I mean, I calculated it two different ways and go the same answer for my test problems.)

Issue #936, opened 2018-11-13
6.5.4 runs FCI when BPC is selected.


cg09:
Unless, as usual, I am mistaken, this seems really bad.  Best to check that the other algorithm selections are actually running the intended algorithms.

jdramsey:
No, this is a general problem. The algorithm selections don't always "stick". If you select an algorithm, run it, then go back to the algorithm selection window, you often get a different algorithm. Same for simulations.

espinoj:
@zhou Can you please look at this problem

yuanzhou:
I thought we added the "Remember" feature when @kvb2univpitt worked on the algo chooser for the 2nd round. Let me check this.

yuanzhou:
I just verified the 6.5.4 branch as well as the development branch, the algo chooser worked well. I tried a few times by selecting BPC for a search. Then I closed the search graph result window, reopened it and clicked the Back button to see the selected algorithm as well other settings, it still says BPC.

kvb2univpitt:
I also tested but couldn't reproduce the problem.

espinoj:
I was able to reproduce.  @kvb2univpitt and @yuanzhou 

graph -> simulate -> search
create a y structure in graph
simulate using SEM
search using GFCI
edit graph (Add a node)
run search since tetrad warns you to
edit graph (add a node)
look at search - it ran BPC!

yuanzhou:
Actually I just reproduced this in a simpler workflow:

Simulation with all default settings -> Search using GFCI
Back to simulation box and change the number of measured variables, Tetrad will ask if to Execute the changes down. Click "Execute" button and the search result will be updated. Click back button to the algo chooser, and BPC is selected.

jdramsey:
I've seen stuff like this happening but couldn't figure out how to reproduce it. Thanks!

yuanzhou:
So I did some debugging and found out that once we click the "Execute" button in the popup warning window, a new search being executed in a new thread, causing the `GeneralAlgorithmRunner` loses the stored models. Because when we reopen the algo chooser, it first resets all the selections then restores from the stored models in `GeneralAlgorithmRunner`. Due to the reset we would see "BPC" being selected since it's the first one in the algo list.

@kvb2univpitt we may need to think about a new way of remembering user selections. I remember seeing Joe's using "Preferences.userRoot()" in other places for storing user data. Or the user selections should be stored in a different place other than "GeneralAlgorithmRunner".

[...]

PR #954, opened 2018-12-03
Algo chooser restore states zhou

yuanzhou:
This pull request addressed the issue #936 

In short this behavior would happen if we make changes to the upstream nodes. For example we did a search from a simulated data, then we change the simulation. Because a new algo runner is created, we lose the user selections on the algo chooser UI.

When the data type is changed in upstream nodes, Tetrad will popup a message window to ask users to run the search again. In this case, all user selections are rest to the default.

When changes can still be populated to the search box and a new search with the same algorithm (test & score) created, we could still restore the selected algo name (from the search result) in UI to avoid any confusions. But the selected test and score are not being restored in this fix, due to inconsistency of algorithm implementations. We'll work on that after the release.

@espinoj @kvb2univpitt @jdramsey please review. Thanks!


espinoj:
I tried to replicate the bug and it seems fixed.  If more people can try I'll merge.

kvb2univpitt:
Just tried.  It works as expected.

PR #955, opened 2018-12-03
Updated Bootstrapping Tables with pd/dd nl columns

chirayukong:


yuanzhou:
I tested with a simulation->search->graph, the bootstrap resulting table stays there in the search after adding a graph box or revisiting the search box.

But the table result disappeared using the "Simulate, Search, Compare" pipeline, after double clicking the Compare box and choosing "Graph-Graph Comparisons". The bootstrap table stays there when I chose "Simulation-Algorithm Comparisons".

PR #959, opened 2018-12-10
Correcting algorithm type for Multinomial Logistic Regression Wald (single line change).

jdramsey:


jdramsey:
This is is a simple one-line fix. Dan Malinsky's colleague noticed the Multinomial Logic Regression Wald test would not run with mixed data, the base case. The reason was it was marked as for continuous data, not mixed. Switch the marking.

PR #964, opened 2019-01-08
Fixes the problem of scatterplots being slow to render.

jdramsey:


jdramsey:
This fixes the problem of scatterplots being slow to render. To test, load a dataset with more than 5000 samples and call the scatterplot menu item. Try switching variables. You will see it renders quickly now. The problem was an expensive method being called in the middle of a loop.

jdramsey:
More detail: If you load a dataset with say 5000 or more records in the current version and pick a variable and try to call the scatterplot item in the menu item, there is a significant (very long) delay before displaying the histogram. If you change variables, you experience also a very long delay. If you close the dialog, another very long delay. This should fix that. With this change, the scatterplots should render instantaneously. If you change variables, the histograms should also be instantaneous. If you close the dialog, the dialog should close instantaneously.

PR #965, opened 2019-01-09
Ranges on a couple of the PC params were wrong. Fixed.

jdramsey:


jdramsey:
More detail: The ranges I changes are obvious fixes. There are three options for each, but the range given was 1 to 1, just one option. For instance, you could you specify CPC in the interface; that option was blocked.

PR #983, opened 2019-03-04
Intervention & metadata handling

yuanzhou:
This pull request contains the following changes:

1. Handling dataset that contains interventional variables (status and value) via the updated data-reader v1.1.0. Updated the spreadsheet-like data editor with headers notions to indicate variable types.

2. Added metadata reading in data loader for overwriting variable data type. Especially useful for interventional data to specify and status and value pairs as well as the data types.

3. Added Determinisim for detecting and merging deterministic interventional variables. Fully determinised domain variables will be detected as well and marked via Node attribute.

4. Knowledge box editor changes for putting all interventional variables to top tier and domain variables in the second tier. As well as generating forbidden group based on required group.

5. Highlight and hide/show interventional variables in a search resulting graph box.

6. Incorporated @kvb2univpitt's fix to concatenating multiple datasets from the first round.


cg09:
Jeremy,

New stuff should not go into TETRAD without documentation. Do you have any
for the intervention modifications? Could you send it to me please if you
do, and if you do not, maybe write some, or get someone to, again please?

Thanks,

Clark

espinoj:
@cg09 Yes.  We will not make the release announcement until the documentation is done.  Zhou is working on it.  @yuanzhou

PR #987, opened 2019-03-06
Fixed list not thread-safe

chirayukong:


yuanzhou:
I'll leave this to @kvb2univpitt since I don't know much about java threading.

Issue #991, opened 2019-03-13
Multiple Graphs Not Shown for non-MultiDataSetAlgorithm.


kvb2univpitt:
When running FGES, a non-MultiDataSetAlgorithm, on multiple imported datasets, multiple graphs (one graph for each dataset) are not shown.

kvb2univpitt:
Fixed.  See branch v6.7.0-multiple_graph_display

![Screenshot from 2019-03-25 16-36-50](https://user-images.githubusercontent.com/10729309/54952617-cc7e0400-4f1c-11e9-893a-b4a8e99a9f06.png)

jdramsey:
Thanks!

PR #1002, opened 2019-03-25
V6.7.0 multiple graph display

kvb2univpitt:
A fix for issue #991.


yuanzhou:
Just tested and it looks good in the search resulting graph panel. But when I attached a graph box to the search box, I only see the first one graph instead of all graphs. Prob an easy fix in the graph editor. Thanks!

kvb2univpitt:
Oh I see.   The problem is in two places.  Fixing it now.  Good catch!

kvb2univpitt:
@yuanzhou It looks like in the previous Tetrad multiple graphs were also not shown in the graph box.  From looking at the code, the GraphEditor is meant to handle only one graph.  Let's wait and see what @jdramsey says about making the GraphEditor showing multiple graph.  This will be a new issue since we are adding new feature not bug fixing.

yuanzhou:
I agree with @kvb2univpitt. Currently it seems the graph box is being used for two purposes:

1. Create a graph box and make new graph, then use the graph with PM, IM and do a search using the graph as input.

2. Connect a graph box to a search box to make changes to the search resulting graphs. This is where the issue comes from. 

We first need to clearly define the role of this graph box and how to use it correctly.

PR #1009, opened 2019-03-29
Changing default for maxDegree to 100 (was 4).

jdramsey:
I just changed this one (very important) default.


yuanzhou:
Due to the changes to `ParamDescriptions`, we no longer define the default value, lower bound, and upper bound in the java class. Instead, you can specify them in the HTML manual.

https://github.com/cmu-phil/tetrad/blob/development/docs/manual/index.html#L2776

Still need to do a `mvn clean package` so the updates get included in the jar.

PR #1010, opened 2019-03-29
Allow Bayes and SEM estimators accepting IM directly

chirayukong:


chirayukong:
We can now connect estimators with IMs.
Screen Shot 2019-03-29 at 3 41 28 PM

PR #1016, opened 2019-03-31
Adjusting alg/test/score names as recommended in https://github.com/cmu-phil/tetrad/issues/1008. 

jdramsey:


jdramsey:
Adjusting alg/test/score names as recommended in https://github.com/cmu-phil/tetrad/issues/1008. Removing any algorithms/test/scores that were recommended (and not already removed). Only discrepancies from that proposal now are Mimbuild and the two IMaGES algorithms (which need to be consolidated).

PR #1017, opened 2019-04-01
Commented a classifier out of the editor toolbar

chirayukong:


yuanzhou:
@jdramsey according to your google doc:

> Evil Classify Box
 Gripe: This needs to be removed entirely. See previous discussion.
Gripe: the Classify box needs to be removed from the manual as well.

Should we totally remove it from the configuration XML files and delete all related java classes as well as the section in HTML manual?

cg09:
Please leave all the Java code for CLASSIFY but delete the interface box.
Madelyn and Clark are going to submit manual changes through Joe, who will
make them directly.  We think this will speed things up, and we can talk
through any issues directly with Joe.

Clark and Joe

yuanzhou:
Based on the above comments, I commented out the code in config xml as well as the html manual.

PR #1019, opened 2019-04-02
Disabled BPC annotation and HTML doc

yuanzhou:
Just commented out the BPC annotation in interface and the HTML manual.


yuanzhou:
#1008

PR #1022, opened 2019-04-02
Removed Whitener data conversion.

kvb2univpitt:
Changes for issue #1021.


cg09:
Why removed?

jdramsey:
My idea. First, it wasn't implemented correctly. Second, all of the algorithms that require whitening already do it internally (the main one being FastICA).

PR #1045, opened 2019-04-06
broken.bayesim.problem

jdramsey:
Fixes four problems.

### DIFF
--- a/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/BayesImEditor.java
+++ b/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/BayesImEditor.java
@@ -79,10 +79,12 @@ public class BayesImEditor extends JPanel {
                 comp.addItem(i + 1);
             }
 
+            comp.setSelectedIndex(wrapper.getModelIndex());
+
             comp.addActionListener(new ActionListener() {
                 @Override
                 public void actionPerformed(ActionEvent e) {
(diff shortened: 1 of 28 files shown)


jdramsey:
This fixes for problems. First, two issues Kong sent regarding estimating models from Simulation->Bayes IM, first with one Bayes model in the simulation box and then with multiple models in the simulation box.

In addition, two simple issues were fixed, for FGES returning a copy of the graph as recommended to avoid problems with edges hanging around when nodes are deleted in the Search box, and second, globally replacing DagToPag with DagToPag2, at Bryan's suggestion, and deprecating DagToPag.

cg09:
Once the fixes are done (?) we need to retest the whole program (ouch!). We
have had a lot of experience with fix X, break Y.

jdramsey:
@cg09 I don't think these changes in particular will have global implications. The fixes for Kong's Bayes IM bugs are specific to the interaction between Bayes IM, Simulation, and Comparison. The fix for Bryan's DagToPag issue--well, all of the unit tests still pass, so I think that's good; the global implications were checked. Returning a copy of the graph from FGES is innocuous but solves a problem I was having editing graphs in the output of of FGES. Of course, @kvb2univpitt, if the graph display for search were rendered properly non-editable, to prevent me from taking advantage of the fact that I can delete variables in that window, this would be a non-issue, so maybe it's an unnecessary fix.

Issue #1046, opened 2019-04-06
IMaGES Continuous and IMaGES Discrete Needs To Be Filtered By Datatype


kvb2univpitt:
Both IMaGES Continuous and IMaGES Discrete should not show up in the algorithm list for either discrete or continuous data set.

![algo_list](https://user-images.githubusercontent.com/10729309/55671855-7be2a100-5862-11e9-8120-68841de46b0b.png)

PR #1047, opened 2019-04-06
V6.7.0 algo datatype

kvb2univpitt:
A fix for issue #1046.

### DIFF
--- a/tetrad-gui/src/main/java/edu/cmu/tetradapp/ui/model/AlgorithmModels.java
+++ b/tetrad-gui/src/main/java/edu/cmu/tetradapp/ui/model/AlgorithmModels.java
@@ -20,6 +20,7 @@ package edu.cmu.tetradapp.ui.model;
 
 import edu.cmu.tetrad.annotation.AlgType;
 import edu.cmu.tetrad.annotation.AlgorithmAnnotations;
+import edu.cmu.tetrad.data.DataType;
 import java.util.Collections;
 import java.util.EnumMap;
 import java.util.LinkedList;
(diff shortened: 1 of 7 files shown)


tetrad-gui/src/main/java/edu/cmu/tetradapp/ui/model/AlgorithmModels.java
```diff
@@ -69,13 +70,28 @@ public static AlgorithmModels getInstance() {
...
 
-    public List<AlgorithmModel> getModels() {
-        return models;
+    private List<AlgorithmModel> filterInclusivelyByAllOrSpecificDataType(List<AlgorithmModel> algorithmModels, DataType dataType) {
+        return (dataType == DataType.All)
+                ? algorithmModels
+                : algorithmModels.stream()
+                        .filter(e -> {
+                            for (DataType dt : e.getAlgorithm().getAnnotation().dataType()) {
+                                if (!(dt == DataType.All || dt == dataType)) {
```
yuanzhou: I think the logic here in the for loop is not right. For an algorithm that can handle both continuous and discrete data, if the annotation has continuous specified before discrete, and the input data is discrete, this algorithm won't be included. For example. 

````
@edu.cmu.tetrad.annotation.Algorithm(
        name = "GFCI",
        command = "gfci",
        algoType = AlgType.allow_latent_common_causes,
        dataType = {DataType.Continuous, DataType.Discrete, DataType.Mixed}
````

GFCI won't show up in the algo chooser when I use some simulated discrete data, because the first data type is continuous and it will return false.

kvb2univpitt: Good catch!  Just fixed it.

yuanzhou:
#1008

Issue #952, opened 2018-11-30
simulation saved output incorrect


rubenSaro:
Using the Tetrad GUI, if you use the Simulation box to create, let's say 10 simulations from _different_ graphs (forward DAGs, structural equations), and then save the simulation, the txt files show different graphs than the ones shown in the Simulation window "True Graph".

Issue #617, opened 2017-12-05
Eliminate load simulation option in simulation databox


espinoj:
Eliminate load simulation in the simulation databox n file menu of simulation box
Rephrase "Save Simulation" to "Save Simulated Data" in file menu of simulation box

related to #616

jdramsey:
This is fixed.

PR #1052, opened 2019-04-08
Fixes the problem with Save Simulation in the Simulation box menu.

jdramsey:
Save Simulation was saving out random datasets, not the ones in the Simulation box. This fixes that problem by creating a new save method in Comparison for just a single simulation. Also, the load graph text variable separator was set to semicolon (;) as in the current save graph text. A unit test was fixed for that.

### DIFF
--- a/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/SimulationEditor.java
+++ b/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/SimulationEditor.java
@@ -355,13 +355,13 @@ public final class SimulationEditor extends JPanel implements KnowledgeEditable,
             if (selectedFile == null) {
                 return;
             }
-            new Comparison().saveToFiles(selectedFile.getAbsolutePath(), simulation.getSimulation(),
+            new Comparison().saveToFilesSingleSimulation(selectedFile.getAbsolutePath(), simulation.getSimulation(),
                     simulation.getParams());
         });
 
-        file.addSeparator();
+//        file.addSeparator();
 
-        file.add(loadSimulation);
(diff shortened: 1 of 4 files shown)


jdramsey:
Resolves this issue: https://github.com/cmu-phil/tetrad/issues/952

Also, I commented out the Load Simulation item. I tried valiantly to get it to work, with some success, but it's still a long way from being useful. For instance, it can't load discrete data. For continuous simulations it works fine. But I agree, we shouldn't advertise falsely. 

Resolves this issue: https://github.com/cmu-phil/tetrad/issues/617

jdramsey:
The way to test this is to make a simulation in the Simulation box, then select Load Simulation, select a directory to put the saved simulation, save. Then go and look at the dataset(s) and graph(s) in the saved simulation and compare them to the ones in the Simulation box. They should be the same.

I'm not sure who should review this.

yuanzhou:
I'm looking at and testing the fix now

yuanzhou:
Tested and compared with simulating single and multiple discrete/continuous data and it looks good. Only thing I'm not sure if the `parameters.txt` says something like this instead of the simulation parameters in simulation setup:

Example 1: "Bayes net simulation using Graph constructed by adding random forward edges"

Example 2: "Linear, Gaussian SEM simulation using Graph constructed by adding random forward edges"

jdramsey:
Oh you're right! Hold on. I hadn't paid attention to parameters.txt. Give me a second.

jdramsey:
The problem is none of the parameters are being marked as used, or if they are the information is being forgotten. I'll think about that. It's a separate issue though, I think. What's in the parameters.txt file is correct so far as it goes.

PR #1055, opened 2019-04-08
Fixes the problem with parameters not being saved out when saving simulation.

jdramsey:
Fixes the problem with parameters not being saved to the parameters.txt file when saving a simulation.


jdramsey:
@yuanzhou I got it! What I said wasn't the problem at all; I had just commented out the wrong line. I cleaned up the code, and now it works. The check it, make a simulation, and save it using the Save Simulation menu item, then look at the saved simulation. It should be the case that the parameters used in the simulation all appear there. All that stuff was working perfectly.

Issue #1013, opened 2019-03-29
A data box will not accept correlation data in another data box 6.7.0-0.


cg09:

jdramsey:
Good catch.

jdramsey:
Fixed.

PR #1056, opened 2019-04-08
Fixes the problem where correlation matrices wouldn't copy over into child Data boxes.

jdramsey:
https://github.com/cmu-phil/tetrad/issues/1013


jdramsey:
To test, make a continuous simulation, S, and make the graph S->Data1->Data2 in the session workspace. For Data1, convert S to a correlation matrix. For Data2, just put a Data box. The correlation from Data1 should now appear in Data2. It wasn't doing that before.

@kvb2univpitt

PR #1059, opened 2019-04-09
sundry.fixes.joe

jdramsey:
Fixing two small issues, as indicated.


jdramsey:
Somehow these changes got lost; I'll make them again. The first issue is that you could make graphs containing 2-cycles and the Graph Properties action would deem them acyclic. I'm putting back the fix for this. Evidence that it works is in this picture:

Screen Shot 2019-04-10 at 7 33 06 PM

In general, to test it, you should be able to make any graph containing a 2-cycle, select Graph Properties from the menu, and it should correctly say that the graph is cyclic.

The logistic regression fix simply throws an exception of the dataset is not a mixture of binary and continuous variables. See the following two images. You can test this by mixed datasets with all binary discrete variables or arbitrary discrete variables and see if the error message box pops up.

Screen Shot 2019-04-10 at 7 37 33 PM

Screen Shot 2019-04-10 at 7 37 57 PM

jdramsey:
A picture of what should happen if you have a dataset that's (correctly) a mixture of binary and continuous variables. It should launch the Logistic Regression editor.

By the way, I was just trying clarify the issue; if there are other changes that need to be made, I'm not opposed.

Screen Shot 2019-04-10 at 7 37 12 PM

PR #1071, opened 2019-04-12
Adjusted parameter defaults, plus a couple of other things.

jdramsey:
Reset defaults parameters as suggested in https://github.com/cmu-phil/tetrad/issues/1065.

Resequenced the list of simulations in SimulationEditor to match those in the dropdown there. (A couple were wrong.)

### DIFF
--- a/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/SemImParamsEditor.java
+++ b/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/SemImParamsEditor.java
@@ -108,7 +108,7 @@ public class SemImParamsEditor extends JPanel implements ParameterEditor {
             }
         });
 
-        final DoubleTextField covLowField = new DoubleTextField(params.getDouble("covLow", 0.1),
+        final DoubleTextField covLowField = new DoubleTextField(params.getDouble("covLow", 0.0),
                 6, decimalFormat);
 
         covLowField.setFilter(new DoubleTextField.Filter() {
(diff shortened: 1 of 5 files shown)


jdramsey:
@cg09 This pull requests resets all of the default parameter we talked about to better default values, plus some. You won't be able to check them unless I send you a special build, but if you want, let me know, I'll make one.

jdramsey:
The only other thing I did here was to change the non-Gaussian option in the linear Fisher model to abs(Normal) instead of Beta, so those parameters were all adjusted, as well as the underlying code. I think this is more workable default non-Gaussian distribution.

The idea is that if you use the linear Fisher method of simulation, you can select abs(Normal) to be the distribution of the errors and get a very easy linear, non-Gaussian simulation, without all of the work of making a generalized SEM PM, IM, etc.

@cg09

jdramsey:
https://github.com/cmu-phil/tetrad/issues/1065

tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/SemImParamsEditor.java
```diff
@@ -108,7 +108,7 @@ public double filter(double value, double oldValue) {
             }
         });
 
-        final DoubleTextField covLowField = new DoubleTextField(params.getDouble("covLow", 0.1),
+        final DoubleTextField covLowField = new DoubleTextField(params.getDouble("covLow", 0.0),
```
yuanzhou: Instead of hardcoding the default value, can you get it by calling `ParamDescriptions`?

````
params.getDouble("covLow", (Double) ParamDescriptions.getInstance().get("covLow").getDefaultValue())
````

jdramsey:
The problem is then you can't have two IM's with different settings.

tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/SemImParamsEditor.java
jdramsey: The problem is then you can't have two IM's with different settings.

yuanzhou: Oh I see your point.

Issue #1072, opened 2019-04-12
Make all graphs non-editable in the Tetrad GUI except for those displayed by GraphEditor


jdramsey:
@kvb2univpitt I went through the editors in the Tetrad GUI. I think the rule is very simple. All graphs should be non-editable, except for those that use the GraphEditor as an editor. So I think if graphs were by default non-editable, but in the GraphEditor editing were turned on, that would be completely satisfactory.

I'll have to get used to not be able to delete nodes in the search box, but I will adjust. It's for the best. :)

kvb2univpitt:
@jdramsey The rule makes sense.

PR #1073, opened 2019-04-14
V6.7.0 disable graph editing

kvb2univpitt:
Fixes for issue #1072.


kvb2univpitt:
@jdramsey I think I've covered it all.

PR #1081, opened 2019-04-16
V6.7.0 rework algo editor

kvb2univpitt:
edu.cmu.tetradapp.editor.GeneralAlgorithmEditor is getting too big to manage.  Separated out GUI components.


yuanzhou:
The back and forward buttons are disabled when I didn't enter the target variable name. After I entered, those buttons are still disabled so I couldn't proceed unless clicking the Done button and do it again.

Capture

kvb2univpitt:
@yuanzhou I purposely disable those buttons so that the users can't execute multiple run.  I need to fix it so that they are enabled when an error occurs.  Good catch!

kvb2univpitt:
@yuanzhou Ok.  It's fixed.

yuanzhou:
Another behavior I noticed is when I started with an algorithm say FOFC, I changed the alpha parameter to 0.02, then if I went back to choose a different algorithm say GFCI, instead of seeing the alpha being reset to 0,01, it was still 0.02. Similarly, if two algorithms share some common parameters, when switching from one algorithm to another one, the parameter values set earlier won't be reset for the new chosen algorithm. I verified with the development branch and it does reset the values.

PR #1089, opened 2019-04-18
V6.7.0 filter algo on dataset

kvb2univpitt:
Algorithms are now also filtered by number of datasets.

Single dataset:
single

Multiple datasets:
multidataset

Also, since only one graph is produced, we got rid of the tab panel:
graph


jdramsey:
Nice!

yuanzhou:
Tested and looks good.

Issue #1090, opened 2019-04-19
Suggested defaults for discrete searches


jdramsey:
Based on simulations with variables with 3 categories using the algorithm comparison tool, some recommended defaults:

* Alpha for discrete variable searches with PC algorithms = 0.05

   - This is apparently different from continuous searches, so need a separate parameter under the hood.

* Collider discover rule for PC searches = CPC

  - For very small searches, PC-Max

* Conflict rule for PC searches = prioritize existing colliders

* Sample prior for FGES/BDeu searches = 20

* Structure prior = depends on the density but a default might be 3

jdramsey:
Note that our current defaults for sample prior and structure prior for FGES/BDeu lead to some very bad searches, so this should be changed soon. On it.

jdramsey:
Looking at models with 2-4 categories randomly across variables. The advice holds up pretty well. Maybe a default of 15 instead of 20 for FGES/BDeu is better. Also, PC-Max is better for sample size 100 as well, so the advice should be that for small sample sizes PC-Max is better; for sample sizes > 100 CPC. (I checked sample size 200--CPC was better.)

jdramsey:
I left out one search option--FGES/Discrete BIC. I ran a comparison of FGES/BDeu against FGES/Discrete BIC. Usually the BDeu option is better, with the default sample and structure priors above, but discrete BIC is almost as good. The best penalty discount for it is clearly 1 (every time). Maybe we should just remove the penalty discount option for this.

jdramsey:
Also, penalty discount for discrete BIC should according to all of my result always be set to 1 (standard BIC). We can effectively remove penalty discount as a parameter for this score.

cg09:
what does "very small" searches mean--variables or samples and how small,
roughly of course.

jdramsey:
@cg09 I mean by "very small" sample sizes of 100 or less. I know not everything thinks that's small, but from the point of view of our searches it definitely is.

Issue #1087, opened 2019-04-17
Suggestions for discrete simulation and search.


jdramsey:
These suggestions (revision):

a. Switch the distribution for ML Bayes IM for coefficients from Uniform to Beta.

b. Standardize the following across simulation method to the extend possible:

   * Discretization
   * Standardization
   * Random column reordering

c. Compare our chi square calculations to those in R.

jdramsey:
For (b), there may be a second requirement from the algorithm comparison tool. These parameters may need to be included in the simulations, not sure.

jdramsey:
I think all of this is done except for comparing our chi square calculations to R. I'll do that this weekend.

jdramsey:
All done.

PR #1092, opened 2019-04-19
Adjusted some defaults for discrete searches to optimize performance.

jdramsey:
https://github.com/cmu-phil/tetrad/issues/1090

### DIFF
--- a/docs/manual/index.html
+++ b/docs/manual/index.html
@@ -2982,7 +2982,7 @@ there is a single latent for each output cluster. For MimBuild, a pattern over t
         <h3 id="alpha" class="parameter_description">alpha</h3>
         <ul class="parameter_description_list">
             <li>Short Description: <span id="alpha_short_desc">Cutoff for p values (alpha) (min = 0.0)</span></li>
-            <li>Long Description: <span id="alpha_long_desc">Statistical tests often compare a test statistic to a distribution and make a judgment that the null hypothesis has been rejected based on whether the area in the tails for the distribution for that test statistic is greater than some cutoff alpha. For tests of independence, for instance, a lower alpha level makes it easier to judge independence, and a higher alpha makes it harder to judge independence. Thus, a lower alpha for a search generally results in a sparser graph.</span>
+            <li>Long Description: <span id="alpha_long_desc">Statistical tests often compare a test statistic to a distribution and make a judgment that the null hypothesis has been rejected based on whether the area in the tails for the distribution for that test statistic is greater than some cutoff alpha. For tests of independence, for instance, a lower alpha level makes it easier to judge independence, and a higher alpha makes it harder to judge independence. Thus, a lower alpha for a search generally results in a sparser graph. The default for this is 0.01, though for discrete searches we recommend using a value of 0.05.</span>
             </li>
             <li>Default Value: <span id="alpha_default_value">0.01</span></li>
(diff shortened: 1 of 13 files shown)


jdramsey:
@yuanzhou I edited the manual a bit. I couldn't figure out a good way to give the discrete searches different defaults for alpha so I just noted in the manual that for discrete searches the default should be 0.05 for Chi Square and G Square. Otherwise, as in https://github.com/cmu-phil/tetrad/issues/1090.

jdramsey:
@yuanzhou I standardized the simulation parameters mentioned in https://github.com/cmu-phil/tetrad/issues/1087. Now it should be the case that:

- The following have post-processing discretization parameters.

   * None. The only simulation methods with discretization parameters are the mixed variable simulations, Lee & Hastie and Conditional Gaussian.

- The following methods should have a "standardize" parameter:

   * SEM Simulation
   * Linear Fisher
   * Time Series Simulation

- The following methods should have a method to randomly reorder columns:

   * All but Time Series Simulation.

jdramsey:
@yuanzhou Please don't merge this quite yet. Some further issues from Clark.

jdramsey:
@yuanzhou OK, this is good I think. I wrote a script to help find defaults for the parameters for discrete searches, which I'm committing with this pull request. I set the priors to the ones listed. I played around a lot with discrete searches, but I think I put everything back to the way it was, except where noted.

yuanzhou:
Looking at this now

yuanzhou:
@jdramsey can you please describe couple of the testing cases so I know where to look at and don't miss any intended uses? I looked at the code but was not clear about the test cases.

jdramsey:
Well, I mean, what you could do would be to re-run the script I included for, say, sample size 500 and see if you agree with the defaults I picked. They should be for models that occur near the top the output under the Average Statistics section. One of the pushes I did was for that script. All you need to do is run the script; it should be automatic.

yuanzhou:
[Discrete_Searches_Sample_Size_5000.txt](https://github.com/cmu-phil/tetrad/files/3109405/Discrete_Searches_Sample_Size_5000.txt)

[Discrete_Searches_Sample_Size_500.txt](https://github.com/cmu-phil/tetrad/files/3109406/Discrete_Searches_Sample_Size_500.txt)

@jdramsey Here are the results I got for sample size 5000 and 500. I can merge this pull request if they look good. You much have a better idea about the defaults.

jdramsey:
Sorry, there is a file for each that would have appeared in your tetrad directory. Do you see them?

jdramsey:
They're in a folder called comparison.discrete.* something. Those contain all of the rankings of the models.

yuanzhou:
Ohh, I was wondering about the Average Statistics section you mentioned but didn't see them in the output. 

Here are the rankings:


[comparison_all_5000.txt](https://github.com/cmu-phil/tetrad/files/3109497/comparison_all_5000.txt)
[comparison_all_500.txt](https://github.com/cmu-phil/tetrad/files/3109498/comparison_all_500.txt)

jdramsey:
Perfect. Your N = 500 came out even better than mine:

jdramsey:
I guess I can also say, if make a random Bayes net with 3 categories per node, average degree 2, and search it using FGES/BDeu or PC with the CPC option, with these defaults, you should get a result which is very close to the pattern of the true graph. That was the idea.

Just giving recommendations to people as to what algorithms and settings to use.

Issue #1107, opened 2019-04-30
Only show edge type probabilities columns when there's a bootstraping


yuanzhou:
Currently, in the Graph box, we have a table under the graph that shows the edges and edge type probabilities. Those edge type probabilities columns will confuse users since there's no bootstraping. 

![Capture](https://user-images.githubusercontent.com/195873/56976016-0192fd00-6b40-11e9-9c86-bc4047d54974.PNG)

And similarly,  in the search resulting graph panel.

![Capture](https://user-images.githubusercontent.com/195873/56976056-1c657180-6b40-11e9-8d8d-eee88c29605c.PNG)

Since we can also connect a graph box to a search box, we should only show the edge type probabilities columns when the search has bootstraping.

kvb2univpitt:
Fixed.  Also, add stripped rows for easy reading.

Without Bootstrap:
![no_bootstrap](https://user-images.githubusercontent.com/10729309/57089438-71d18800-6cd2-11e9-9a0c-1295994de985.png)

With Bootstrapping:
![boostrap](https://user-images.githubusercontent.com/10729309/57089484-8746b200-6cd2-11e9-84a7-fa0db357e967.png)

jdramsey:
Nice.

yuanzhou:
I'm looking at the fix now, thanks @kvb2univpitt

PR #1108, opened 2019-05-02
V6.7.0 bootstrap table

kvb2univpitt:
A fix for issue #1107.


yuanzhou:
Looks beautiful! My only suggestion is to make the table a little shorter by default and make the graph area taller.

Captur22e

Capture

kvb2univpitt:
OK.  I'll work on your suggestion.

kvb2univpitt:
@yuanzhou I adjusted the split panel divider location.

Issue #1109, opened 2019-05-02
Some issues getting verbose output to work for algorithms.


jdramsey:
Verbose output should work for the following algorithms if the verbose flag is set to true (some of these already work):

1. PC (already works)

2. FCI (already works)

3. FGES 

- I just adjusted the FGES class for this--it was still using the old logger methods. Need to push it.

4. FOFC

- The pure tetrad should be printed out. This used to work.

Also, the verbose flag should alway occur LAST in the list of parameters. Just need to go through and make this the case.

Also, make this version available to @cg09 and @ps7z so they can use it.

@cg09 @ps7z @mglymour Did I get everything?

jdramsey:
I marked some algorithms as for continuous data only: R3, Skew, RSkew, LiNGAM, GLASSO.

jdramsey:
Just a note: In a couple of cases, the verbose parameter was coming up from another source and had to be removed from the list and re-added to get it to come out as last.

This all works now for me (I think). OK, I'll do the pull request.

yuanzhou:
Addressed via #1110

PR #1110, opened 2019-05-02
Fixed verbosity issues, "verbose" last issue, data type for some continuous searches.

jdramsey:


jdramsey:
https://github.com/cmu-phil/tetrad/issues/1109

jdramsey:
To test the verbosity issues, run searches using PC, FCI, FGES, FOFC, with logging turned on in the interface. Verbose output should appear (independence test info, score into, etc.)

To test "verbose" last issue, just go through the algorithms in the interface and make sure that no matter which algorithm you pick, the "verbose" option comes out last. (Maybe there's a better way to do this, come to think about it, than just checking it on an algorithm by algorithm basis.)

For the data type issue, make sure the following algorithms are filtered out for discrete and mixed data: R3, Skew, RSkew, LiNGAM, GLASSO.

jdramsey:
Also, if this is accepted, could you please let me know how Peter and Clark can get a build with these changes?

yuanzhou:
I just tested verbose logging, worked fine with PC, FCI, FGES. But when I ran FOFC on continuous data, I got this error:

Capture

And if I checked "No" for MimBuild, I went through but no logging.

Capture

jdramsey:
I see. That's supposed to take a MIM model. If you make a random graph but select the MIM option you should get a pure measurement model, and it should work. But we need a better error message. Let me see if I can make a better error message.

yuanzhou:
Oh I see. Just simulated the Random One Factor MIM for continuous data and the logging worked fine with FOFC.

jdramsey:
OK, I changed the error message to be something more informative. Is this better?

Screen Shot 2019-05-03 at 11 30 08 AM

jdramsey:
Nice! It works!

yuanzhou:
Much better error message! And the `DataType.continuous` annotation you added in R3, Skew, RSkew, LiNGAM, GLASSO works well for filtering.

yuanzhou:
Just came to my mind that should you also annotate FOFC and FTFC for continuous data only? So we don't see them with discrete and mixed data?

jdramsey:
Great!

Issue #1140, opened 2019-06-03
Fask doesn't properly take an initial graph as input.


jdramsey:
In the GUI you should be able to give FASK an initial graph (by drawing an edge from a Graph box to the Search box, in addition to the Data box). It should use that in place of the FAS search. Ruben complains this isn't happening.

This is very easy to fix, which I will forthwith do. I have a working version of the FASK wrapper in another branch, which I'll just copy over and push.

@rubencmu @yuanzhou @kvb2univpitt

yuanzhou:
Merged via #1141

PR #1141, opened 2019-06-03
A fix for https://github.com/cmu-phil/tetrad/issues/1140

jdramsey:


jdramsey:
Fixing https://github.com/cmu-phil/tetrad/issues/1140.

To test, give FASK a dataset and a graph (with the same variables). It should run (which it wouldn't have done before) and use the adjacencies of the graph supplied as a subset of the output adjacencies. If you turn off corr diff adjacencies for FASK, it should use just the adjacencies supplied.

@rubencmu @yuanzhou @kvb2univpitt

yuanzhou:
I'm testing this now

PR #1149, opened 2019-06-11
V6.7.0 data reader migration

kvb2univpitt:
Migrating the data-reader into Tetrad as a module.


yuanzhou:
I tested and it worked well with the maven build. Only two points:

1. We should remove the Version number in the data-reader README heading since its version will be the tetrad version from now on.

2. We probably don't need the LICENSE file inside the data-reader since the tetrad has a license file in root directory. 

Please address the above and I'll merge.

kvb2univpitt:
@yuanzhou Done.

PR #1150, opened 2019-06-12
Tests scores desc zhou

yuanzhou:
- Added descriptions of tests and scores in HTML manual, now all the tests and scores in GUI have descriptions. Thanks @jdramsey !

- Replaced the hard-coded parameter strings with string constants in tests and scores.


jdramsey:
Oh I missed this. Sorry.

PR #1151, opened 2019-06-12
Correcting IndTestFisherZ to use correlations to calculate partial correlations instead of covariances.

jdramsey:
This is a small correction to the Fisher Z test. Previously we were calculating partial correlations by inverting the covariance matrix. But theoretically, it should be the correlation matrix. This just makes that substitution. The class CorrelationMatrixOnTheFly was added to correspond to the class CovarianceMatrixOnTheFly that we were using.


jdramsey:
To test this, just do a search using Fisher Z, like PC with Fisher Z or CPC with Fisher Z. You should get slightly better answers than before.

kvb2univpitt:
Well, it does seem the skeleton of the searched graph is close to the skeleton of the true graph.

PR #1164, opened 2019-06-18
Put back default parameter description.

kvb2univpitt:


kvb2univpitt:
@jdramsey Fixed!  Good catch!

Issue #1175, opened 2019-07-02
Port new covaraince code over to an accessible branch.


jdramsey:
There is some new, well-tested covariance code (including code from Kevin). This needs to be ported to an accessible branch and cleaned up. If OK, it can be moved to development.

kvb2univpitt:
@jdramsey Since my CovarianceOnTheFly class is really not "on the fly," let's rename it.

jdramsey:
@kvb2univpitt What should we name it? I was calling it "KevinCovariances" but maybe not that. :)

jdramsey:
I'll work on that stuff now. The tricky part (other than naming :)) is to see whether I can get rid of some methods that are only used in one place in the code, to make the API easy.

I've got standard covariances and Kevin covariances, for doubles and floats. The current covariancematrix on the fly I just made wrap the other covariancematrix; we can just get rid of it once I find a way to delete the extra methods in that. @kvb2univpitt

jdramsey:
@kvb2univpitt OK, I'm calling the standard calculation "standard" and yours "alternative". Feel free to rename.

The extra methods I was worried about are never used, so I deleted them. Now CovMatrixOnTheFly just wraps CovarianceMatrix and can be factored into nonexistence. I'll do that in the branch I'm making. I realized your calculation even without the parallelization was faster than CovMatrixOnTheFly, so there didn't seem a need for it anymore. I we find a need we can always re-make it, we are that ingenious.

So the basic port will be the following classes:

CovarianceMatrix (recoded)
CovariancesFloatStandard
CovarianceDoubleStandard (please no puns)
CovariancesFloatAlt
CovairnacesDoubleAlt

Also,

CovianceMatrixOnTheFly,

which will be refactored out of the code.

I think there are a few other ways of calculating covariances in the code, which should be replaced by these.

Like I said, I'm just sketching out an API; feel free to adjust it.

Also, when you get your parallelized alt implementation done, you can just add it as a new module. I suggest we keep all implementations around until it becomes clear which are the best performers.

I'll make a branch called joe.covariances.port off of development and move this stuff over.

jdramsey:
This is done.

PR #1176, opened 2019-07-02
joe.covariances.port

jdramsey:
Porting cleaned up covariance classes.


jdramsey:
Here's what I did. I had covariance classes in another branch, including Kevin's implementations. I cleaned these up, put them behind a common interface. I revised the CovarianceMatrix class to calculate covariances using whatever method the user selects, by default Kevin's method for doubles with bias correction. This can be changed. More calculation methods can be added easily.

I changed all uses of CovarianceMatrixOnTheFly to CovarianceMatrix and removed that class.

This incurred a number of unit test failures, which I fixed; they were all for dumb bugs.

Fix for https://github.com/cmu-phil/tetrad/issues/1175.

kvb2univpitt:
Ok.  I am checking it now.

tetrad-lib/src/main/java/edu/cmu/tetrad/data/CovariancesFloatStandard.java
```diff
@@ -0,0 +1,160 @@
...
+
+        for (int i = 0; i < _data.length; i++) {
+            for (int j = 0; j < _data[0].length; j++) {
+                this._data[i][j] = (float) (round(data[i][j] * 100.0) / 100.0);
+            }
+        }
+
+        this.numOfRows = data.length;
+        this.numOfCols = data[0].length;
+        this.covariances = compute(biasCorrected);
```
kvb2univpitt: The compute method should be private if called inside of constructor.  If not, the user can call it again.

tetrad-lib/src/main/java/edu/cmu/tetrad/data/CovariancesFloatStandard.java
```diff
@@ -0,0 +1,160 @@
...
+    @Override
+    public double covariance(int i, int j) {
+        return covariances[i][j];
+    }
+
+    public int size() {
+        return numOfCols;
+    }
+
+    public void setCovariance(int i, int j, double v) {
```
kvb2univpitt: This method should not exist.  A user should not be able to modify the covariance matrix.  This will lead to data corruption.

kvb2univpitt:
@jdramsey I'll work on those issues that I mentioned above from your branch.

jdramsey:
Oh ok. Sorry I was going to get to them. I got busy with another project

kvb2univpitt:
@jdramsey I am going to do  some refactoring.  Basically, I am going to remove CovariancesFloatStandard and just replace with edu.cmu.tetrad.stat.correlation.CovarianceMatrix.  No need to have the same implementation in two different places.  Also, alternative covariance is basically compute the covariance on double data and return float data.  Do we even need an "alternative" covariance?

jdramsey:
I think we need it--alternative covariance is your method.

jdramsey:
Oh, why do we need two? Because I don't know which performs the best yet.

kvb2univpitt:
@jdramsey 

> Oh, why do we need two? Because I don't know which performs the best yet.

Ok.  Now I see the difference.  Well, in that case, I am just going to leave everything the way it is.  I thought it was the same implementation.

kvb2univpitt:
@espinoj I am going to leave everything the way it is.  Do you want to include this change for this coming release?

jdramsey:
Yes, though if you want to add a parallelized version of your algorithm, that would be fine.

Basically, everything in the code should use the same covariance method. That way if we find issues, we can fix them all in one place. That was my idea.

It's OK to leave the choice of covariance method hard-coded for the moment. My issue is, does one of the other lead to better algorithm performance? I still haven't sorted that out, though I spent a fair amount of time on it. I think the double method perform better than the float method. But once this is in development I will do a study comparing the methods all to one another and we will know. Then if we agree, we can remove the methods that are inferior.

Basically, I will use the algorithm comparison tool to run searches for PC and FGES with the different options head to head and see which one does the best. I think it's just heuristic.

jdramsey:
But I'd rather do it while it's in the development branch, just so I don't introduce a bug from my current poison branch.

kvb2univpitt:
@jdramsey I'll add the parallelized version of my implementation to this branch.

kvb2univpitt:
@jdramsey I am going to hold off on merging this into the development branch.  Jeremy is going to make a release this week.  Since we haven't fully tested the new covariance matrices, we won't include it.

jdramsey:
Sounds good! I'll test it in that branch then. @kvb2univpitt

jdramsey:
Canceling this, at Kevin's request.

PR #1197, opened 2019-08-18
Fixed a couple of issues from the issues list.

jdramsey:


espinoj:
I'm guessing is should hotfix this to 6.7?

jdramsey:
That would be helpful. People complain when lingam isn't there, but they
complain more when it doesn't work.

PR #1216, opened 2019-09-24
Degenerate gaussian

chirayukong:


bja43:
Done: I added DG LRT to the degenerateGaussian branch and completed some preliminary testing to make sure it works.

PR #1276, opened 2020-04-15
Added the penalty discount parameter to images--it got deleted somehow.

jdramsey:


jdramsey:
Added a fix for RICF in the SEM Estimator, which was throwing exceptions. The problem there was that some setShowError() calls were misplaced or setting to the wrong things.

jdramsey:
Kevin, The issue here is that the interface through some bug that had been introduced wouldn't actually allow you to use RICF to estimate a MAG, something that worked before. I fixed it.

Maybe we should bet Peter to verify that both the Images and the RICF bugs have been satisfactorily fixed. @ps7z ?

PR #1284, opened 2020-05-05
Added Comparison.compareFromGraphs methods (based on existing methods that were previously debugged, the names of which were changed as per spec).

jdramsey:


kvb2univpitt:
@jdramsey Didn't you say you need to do some more testing on this?  Or, can I go ahead and play around with this?

cg09:
Kevin,

Thanks for keeping at all of this. Really thanks.

Clark

kvb2univpitt:
@cg09 You're welcome.  :)

jdramsey:
Maybe we could talk briefly about what this is supposed to do from the user's point of view? I may not be clear on it.

cg09:
Not sure.

kvb2univpitt:
@jdramsey Ok.  We can talk about it after our meeting on Monday at 2:00 PM.  Thanks.

jdramsey:
Too old to merge....

PR #1295, opened 2020-08-28
joe.ind.test.compare.further.fixes

jdramsey:
A few further fixes. Also, refactored the Chi Square code a little for Clark.

### DIFF
--- a/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/IndependenceFactsEditor.java
+++ b/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/IndependenceFactsEditor.java
@@ -110,8 +110,8 @@ public class IndependenceFactsEditor extends JPanel {
         varNames.add("?");
         varNames.add("+");
 
-        final JComboBox variableBox = new JComboBox();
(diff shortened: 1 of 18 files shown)


jdramsey:
This should work. Notice there is a score-based search (FGES) into the ind facts gadget. If you select show p-value or scores, it shows the scores.

[ind.facts.compare.example.4.zip](https://github.com/cmu-phil/tetrad/files/5142462/ind.facts.compare.example.4.zip)

Also, I cleaned up the chi square code a little. In one spot it was incorrect.

jdramsey:
Also, I went through all of the scores and made them print out their names correctly.

jdramsey:
Also, for algorithms that don't use scores or tests, like LiNGAM, I post an error message. So now you should not be able to produce an unwanted exception using the ind facts editor.

jdramsey:
Clark did some testing of the ind facts gadget. There was one case where he thought it was making a mistake, but in that case the algorithm was making a mistake, not the ind facts editor. So far the ind facts editor (in this latest version) seems to be acting correctly.

jdramsey:
OK, I will now leave this branch alone! :)

PR #1298, opened 2020-09-16
Unicode edge prob table

espinoj:
This is a change to rendering of arrows in the edge prob table from ASCII to unicode so that the arrows don't look like they are dashed.


espinoj:
Reviewed by Greg as an improvement to readibility

PR #1309, opened 2020-10-29
Added warning when using test-wise deletion.

kvb2univpitt:
Added a warning when test-wise deletion is used.


jdramsey (changes requested):
This looks great. I would change the punctuation of the message as follows:

Dataset contains missing values;
testwise deletion will be used.

I like the red.

kvb2univpitt:
Fixed!

PR #1358, opened 2022-01-06
Hotfix 6.9.0

jdramsey:
The GUI was having trouble interfacing with the server for launching. Added an exception catch for this that allows the interface to launch correctly.


jdramsey:
Sorry, I was trying to do the Intellij interface to the pull requests and accidentally committed this. It's well-tested though. Let me know if there's a problem; we can revert the commit.

Issue #1368, opened 2022-03-20
Adjust score/test chooser to give a choice of "Linear/Gaussian", "Mixed Discrete/Continuous", "General", and "All"


jdramsey:

kvb2univpitt:
Fixed.

**With no missing data:**
![no-missing-data](https://user-images.githubusercontent.com/10729309/159214621-9b365308-a638-45d4-b5b0-a28dbdfba945.png)

**With missing data:**
![missing-data](https://user-images.githubusercontent.com/10729309/159214630-a1008d3e-030b-41cf-8d93-f37f78a2d6d3.png)

jdramsey:
Fixed.

PR #1370, opened 2022-03-21
Kevin data filter

kvb2univpitt:
A fix for issue #1368.


jdramsey (commented):
This is much better!! The only issue is that the "Score:" dropdown is hidden now in the default layout. Could the stuff above it be scrunched up? One way to do it I thought of is to take out two of the (obvious) instructions, as indicated in the images below. Is that OK? I commented the out in a hacky way; probably if you like this you should do it though.

Screen Shot 2022-03-21 at 7 47 35 AM
Screen Shot 2022-03-21 at 7 48 56 AM

jdramsey:
Well, that is, "Score:" is hidden on _my_ machine.

jdramsey:
Oh, and obviously I'll need to add a description for FASK-PW, I'll do that. I added several; I thought I'd gotten them all, sorry.

kvb2univpitt:
Ok.  Try it now.  I changed the default size.

Issue #1373, opened 2022-03-24
In Search box, score and test choices do not get preserved in cases where algorithm choices do and all parameters do.


jdramsey:
Branch tetrad-fixes-joe-2022-02-28. This procedure.

Put Simulation-->Search on the workbench.
Double click Simulation, select Structural Equation Model, click Simulation, Done.
Double click Search, select FGES, select EBIC Score, click Select Parameters, Run Search & Generate Graph,  Done.
Double click Search, click Set Parameters, Choose Algorithm, see that EBIC is still selected and FGES is still selected, click Done.
Double click Simulation, click Simulate, click OK, click OK, click Done.
Click Execute.
Double click Search, click Set Parameters, Choose Algorithm, see that FGES is still selected but EBIC Score has now reverted to SEM BIC Score.

kvb2univpitt:
Fixed.

jdramsey:
Thanks!

PR #1376, opened 2022-04-03
Preserve previous selections.

kvb2univpitt:
A fix for issue #1373


jdramsey (approved):
Oh is _that_ how to fix it?! Wow, I never would have thought of that in a million years. It is embarrassing the other ways I thought of that didn't work. It works! Thanks!

PR #1402, opened 2022-08-13
fix ForbiddenGraphModel

lockmatrix:
ForbiddenGraphModel only consider edge in one direction, it is a bug.


jdramsey (approved):
Good idea.

cg09:
What is a "forbidden graph model"?

lockmatrix:
> What is a "forbidden graph model"?

It is a kind of Knowledge Box in tetrad-gui, 
doc: https://htmlpreview.github.io/?https:///github.com/cmu-phil/tetrad/blob/development/docs/manual/index.html#graph_box

Issue #1411, opened 2023-01-04
In edge probability table, put --- directly after -> and <-.


jdramsey:

kvb2univpitt:
Fixed.

![Screenshot from 2023-01-05 01-38-20](https://user-images.githubusercontent.com/10729309/210717603-73e92964-8fe7-4e68-a1d2-8a58fbc13806.png)

```
Graph Nodes:
X1;X2;X3;X4;X5;X6;X7;X8;X9;X10;X11;X12;X13;X14;X15;X16;X17;X18;X19;X20

Graph Edges:
1. X3 --> X19 [X3 --> X19]:0.9406;[X3 --- X19]:0.0594;
2. X16 --> X20 [X16 --> X20]:0.9307;[X16 <-- X20]:0.0198;[X16 --- X20]:0.0495;
3. X6 --> X19 [X6 --> X19]:0.9307;[X6 --- X19]:0.0594;[X6 <-- X19]:0.0099;
4. X9 --> X14 [X9 --> X14]:0.8713;[X9 <-- X14]:0.0594;[X9 --- X14]:0.0693;
5. X1 --- X8 [X1 --- X8]:0.8218;[X1 <-- X8]:0.1188;[X1 --> X8]:0.0594;
6. X1 --- X4 [X1 --- X4]:0.7921;[X1 --> X4]:0.1683;[X1 <-- X4]:0.0396;
7. X10 --> X20 [X10 --> X20]:0.7822;[X10 <-- X20]:0.1683;[X10 --- X20]:0.0495;
8. X2 --> X14 [X2 <-- X14]:0.1683;[X2 --> X14]:0.7525;[X2 --- X14]:0.0792;
9. X1 --> X16 [X1 --> X16]:0.6535;[X1 --- X16]:0.2673;[X1 <-- X16]:0.0792;
10. X3 --- X9 [X3 --- X9]:0.6436;[X3 --> X9]:0.2673;[X3 <-- X9]:0.0891;
11. X5 --> X16 [X5 --> X16]:0.6337;[X5 --- X16]:0.3267;[X5 <-- X16]:0.0396;
12. X8 --- X9 [X8 --- X9]:0.5743;[X8 --> X9]:0.3465;[X8 <-- X9]:0.0693;[no edge]:0.0099;
13. X2 --- X10 [X2 --> X10]:0.1980;[X2 --- X10]:0.5347;[X2 <-- X10]:0.2673;
[...]

PR #1418, opened 2023-01-05
Move undirected edge column next to directed edge column.

kvb2univpitt:
A fix for issue #1411.

### DIFF
--- a/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/EdgeTypeTable.java
+++ b/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/EdgeTypeTable.java
@@ -35,6 +35,8 @@ public class EdgeTypeTable extends JPanel {
 
     private static final long serialVersionUID = -9104061917163909746L;
 
+    private static final Color DARK_GREEN = new Color(0,153,0);
+
     private static final String[] EDGES = {
             "Node 1",
             "Interaction",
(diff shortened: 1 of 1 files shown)


jdramsey:
The table looks wonderful! It's just that the green columns are never used. I'm wondering if the PAG coloring is being applied to the individual PAGs.

Also, the display of the PAG coloring changed (Peter signed off on this). We need to quickly verify which properties are being displayed here. I think the thick green ones should be "visible"--i.e., the property should be NL. These come out as thick blue edges in the display, so maybe that can be made blue? I'm not sure what the thin green ones are. Are these possible latent (property PL)?

jdramsey:
That reminds me, I need to add a menu item for the Graph box to apply or unapply the PAG coloring. I mean to do that. Id you copy a search PAG into another box it loses the PAG coloring and you can't get it back. I'll add this as an issue and assign it to myself to do after yours are all merged.

kvb2univpitt:
Yes, the non-bold green (thin) ones are *possible latent*.  Here's the coloring scheme:

| Edge Type | Edge Property | Display    |
|-----------|---------------|------------|
| <--       | none          | black      |
| -->       | none          | black      |
| <--       | pd,nl         | green      |
| -->       | pd,nl         | green      |
| <--       | dd,nl         | green,bold |
| -->       | dd,nl         | green,bold |

kvb2univpitt:
>  I think the thick green ones should be "visible"--i.e., the property should be NL. These come out as thick blue edges in the display, so maybe that can be made blue?

@jdramsey Can we deal with this as a separate issue?  This issue is about moving the *undirected* column of the table to a different location.

jdramsey:
Yes of course!

jdramsey:
Sorry I got pulled into another meeting that's turned out to last a long
time...

Issue #1426, opened 2023-01-11
Add Row Number to Edge List Table


kvb2univpitt:
It would be helpful to have row-numbered edge table to show the number of edges.  Column sorting should have no affect on the row-numbered column since it is not part of the data.

kvb2univpitt:
Done.

PR #1427, opened 2023-01-11
Kevin table row num

kvb2univpitt:
Enhancement: add row number to edge list table to make it easy to determine the number of edges in the graph.  See issue #1426.

> Note that column sorting does not affect the ordering of the row number, since it is not part of the data.

Screenshot from 2023-01-11 15-03-21

Sorted by interaction:

Screenshot from 2023-01-11 15-03-34

Graph:

Screenshot from 2023-01-11 15-07-57


jdramsey (approved):
Looks good.

PR #1432, opened 2023-01-12
Kevin bs seed

kvb2univpitt:
Added seed to bootstrap parameter.  This is an extension for issue ##1413

### DIFF
--- a/docs/manual/index.html
+++ b/docs/manual/index.html
@@ -7745,6 +7745,25 @@ tabs will not appear.</p> -->
         </ul>
 
         <h3 class="parameter_description"
+            id="seed">seed</h3>
+        <ul
+            class="parameter_description_list">
+            <li>Short Description: <span
+                    id="seed_short_desc">Seed for pseudorandom number generator (min = -1)</span></li>
+            <li>Long Description: <span
+                    id="seed_long_desc">The seed is the initial value of the
+                    internal state of the pseudorandom number generator.</span></li>
+            <li>Default Value: <span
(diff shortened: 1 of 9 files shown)


jdramsey:
This looks really good! For some reason all of the results are coming out better than I'm used to.

jdramsey:
I'll take it!

jdramsey:
OK if I try PC (which I *know* is non-random in every way) successive runs with the same seed yield the same bootstrap probability edge results. Good! So now I take if that the algorithms that don't do that have random elements--I'll have to track those down.

kvb2univpitt:
@jdramsey Please hold off on merging.  I found a bug.  I need to move the random number generator outside of the loop.

kvb2univpitt:
@jdramsey OK.  I fixed it!  The random number generator needs to be created outside of the for-loop or else a new random number generator will be created every iteration of the loop.  Please try it now.  Thanks!

jdramsey:
Oh--is this a bug too? If I set the seed to -1, different runs should come out different but they're all the same...

kvb2univpitt:
@jdramsey I don't think so.  The default (-1) will use the current random number generator **RandomUtil.getInstance()**.  If the seed is > 0, we will use the follow:

```java
RandomGenerator randomGenerator = new SynchronizedRandomGenerator(new Well44497b(seed));
```

Basically, it's the same random number generator but with different seed.

jdramsey (approved):
Great!!

Issue #1434, opened 2023-01-13
Make a new parameter for lambda for the Poisson Prior Score


jdramsey:

jdramsey:
https://github.com/cmu-phil/tetrad/pull/1436

PR #1436, opened 2023-01-13
Added a parameter for lambda for the Poisson Prior Score

jdramsey:
If you run a linear, Gaussian search and select the Poisson Prior Score, there should be a lambda parameter with default 1. The range should be any number > 0.


jdramsey:
https://github.com/cmu-phil/tetrad/issues/1434

kvb2univpitt:
Looks good!

kvb2univpitt (approved):
It works!

Issue #1430, opened 2023-01-12
SP not sensitive to background knowledge


jdramsey:

jdramsey:
https://github.com/cmu-phil/tetrad/pull/1437

PR #1437, opened 2023-01-13
Fixed BOSS and SP to return the correct MPDAG

jdramsey:
Knowledge for SP wasn't hooked up; fixed that. Used the GRaSP getGraph method for both SP and BOSS to fix the MPDAG problem, where if for instance you forbad 3->2 but the method wouldn't return an edge (where appropriate) 2->3.


jdramsey:
https://github.com/cmu-phil/tetrad/issues/1430

bja43 (approved):
looks good to me

jdramsey:
Thanks!

PR #1438, opened 2023-01-13
Added graph type selector for the Edge List Comparison.

jdramsey:
To test, check out this branch, put a Simulation box on the workbench, arrow to a Search box, and arrows from Simulation and Search to Compare. Make linear Gaussian data (SEM) with a fair number of variables and edges and then do a search like PC. Then in the comparison box, select Edge Comparison. There should be a menu to let you select which graph type you want to compare to. Select DAG or CPDAG or PAG; the edges from the comparison graph should change. For DAG they should all be --> edges; for CPDAG they should be --> or --- edges; for PAGs, some edges should have circle endpoints.


jdramsey:
Added some updates to manual, so that should be up to date as well now.

kvb2univpitt (approved):
Did a walk-through with Joe.  Looks good.

PR #1441, opened 2023-01-14
Merge some work on experimental algorithms

jdramsey:
If the 'experimental' flag is selected there should be these three experimental algorihtms: LV-Swap1, LV-Swap2, and LV-Swap3.


kvb2univpitt (approved):
All tests passed!

PR #1443, opened 2023-01-15
Final merge for 7.1.3

jdramsey:
Final changes to merge into development for 7.1.3:

* Fixed a few discrepancies from the last merge.

* Fixed loading of text format graph files with bootstrapping information

* Added menu items ot the Graph box File menu for saving and loading Tetrad graphs in PCALG format.

### DIFF
--- a/tetrad-lib/src/main/java/edu/cmu/tetrad/graph/Edge.java
+++ b/tetrad-lib/src/main/java/edu/cmu/tetrad/graph/Edge.java
@@ -257,15 +257,9 @@ public class Edge implements TetradSerializable, Comparable<Edge> {
             String n1 = getNode1().getName();
             String n2 = getNode2().getName();
 
-            // bug! 2022/12/07
-//            if (n1.compareTo(n2) > 0) {// Sort node's names
-//                n1 = getNode2().getName();
-//                n2 = getNode1().getName();
-//            }
-
             for (EdgeTypeProbability etp : edgeTypeDist) {
                 double prob = etp.getProbability();
-                if (prob > 0) {
+//                if (prob > 0) {
                     StringBuilder _type = new StringBuilder("" + etp.getEdgeType());
                     switch (etp.getEdgeType()) {
                         case nil:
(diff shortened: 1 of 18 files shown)


jdramsey:
To test bootstrapping txt loading, run GFCI with bootstrapping, save the graph out as txt, then load it as txt in a different graph box. First of all, it should load now, and the edge probability table should be the same as the one saved out.

To test the PCALG format saving and loading, save the above graph out as PCALG, then load it as PCALG. You should get the same graph back. (PCALG format doesn't save bootstrapping information.)

jdramsey:
Also, marking of latents in text graph format had been implemented. If you make a graph with latents, save it out in the text format, and load the graph in a different box, the latents should correct.

jdramsey:
Also, there is a menu item in the Graph box that lets you mark a graph as a PAG. If you do this, the PAG coloring will be applied.

bja43:
The parameter description for resampling size is misleading. It says the the min is 0.1, implying that it should accept a range between 0.1 and 1, where 0.1 is 10% and 1 is 100%. However, the default value is 90, which I assume is 90% and if you give it any value less than 1, the parameter appears to be ignored and the last valid value is used instead.

bja43:
- Saving and loading of graphs resulting from bootstrapped searches seems to be functioning property. 
- Saving PCALG graphs seems to be functioning properly.

bja43:
- Non-maximal graphs can be marked as PAGs without an error or warning being thrown.
- Non-ancestral graphs can be marked as PAGs without an error or warning being thrown.
- When a graph is marked as a PAG, clicking an edge changes it's coloring. For example, if you click on the head of a bold arrow, then it becomes a non-bolded dashed arrow. If you unmark as PAG and then remark as PAG, the edge is once again bolded.

jdramsey:
Will fix the errors. As for thrown an exception of non-PAG, one item on my to-do list for the next version is a test for legal PAG, and I can fix that problem then.

But I will fix the parameter range.

I can easily re-apply PAG coloring after each change, will do that.

bja43 (approved):
Looks good!

tetrad-lib/src/main/java/edu/cmu/tetrad/graph/Edge.java
```diff
@@ -307,13 +301,13 @@ public final String toString() {
                     }
                     buf.append("[").append(_type).append("]:").append(String.format("%.4f", etp.getProbability())).append(";");
 
-                }
+//                }
             }
         }
 
-        if (probability > 0.0) {
```
kvb2univpitt, line 314: @jdramsey Why is the edge probability printout code removed?

PR #1449, opened 2023-01-17
Sundry interface fixes

jdramsey:
* Fixed the graph comparison type choosers for the edgewise, misclassification, and stats list comparisons so that they remember the last choice you made.
* In the stats list comparison, moved the new path statistics to the bottom of the list for backward compatibility.
* Made all random graph dialogs cancelable.
* Updated the date on the manual.

Some @expeirmental code was also updated.


kvb2univpitt (approved):
Looks good.

PR #1456, opened 2023-01-19
Cpdag knowledge issue grasp

jdramsey:
The method in TeyssierScorer returning a CPDAG was adding unforced edges into the graph. Switching to a different orientation method. This affects GRaSP, BOSS, SP, etc.


bja43 (approved):
Looks good!

PR #1460, opened 2023-01-22
Legal Pag

jdramsey:
Added test for legal PAG and worked this into the relevant algorithms and into the PAG-labeling menu items throughout the code.

### DIFF
--- a/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/GraphEditor.java
+++ b/tetrad-gui/src/main/java/edu/cmu/tetradapp/editor/GraphEditor.java
@@ -425,6 +425,7 @@ public final class GraphEditor extends JPanel implements GraphEditable, LayoutEd
         graph.add(new JMenuItem(new SelectBidirectedAction(this.workbench)));
         graph.add(new JMenuItem(new SelectUndirectedAction(this.workbench)));
         graph.add(new JMenuItem(new SelectLatentsAction(this.workbench)));
+//        graph.addSeparator();
         graph.add(new PagTypeSetter(getWorkbench()));
 
         menuBar.add(graph);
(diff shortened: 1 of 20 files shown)


jdramsey:
@bja43 Please don't merge this quite yet--need to think some more. Here are the changes I made from the last time you looked at the legal PAG algorithm.

* I added a check in the legal PAG algorithm to make sure all variables in the graph are measured.
* In the menu item in all relevant Graph menus (e.g., in the Graph box and Search box), for the "Set type to PAG" item, if this is to add a PAG designation for the graph, I do a legal PAG check, and if it fails, I warn the user in a dialog that the graph is not a legal PAG and ask if they want to proceed. I only change the type to PAG if they say Yes.
* In all places in the code where a graph was being set to type PAG, I do a legal PAG check, and if it fails, I do not so set the type. The graph is not then colored, and to get the coloring, the user has to Graph menu item to set the type to PAG (with the warning).

The effect has been (is this good? bad?) that a lot of LV algorithms do not have their outputs designated as legal PAGs and so aren't colored. This may be a bit confusing. I'm not sure how to get around the confusion, but at least it has the virtue of not designating graphs as PAGs programmatically that aren't legal PAGs.

It occurred to me that if the program says a graph is not a legal PAG, it might be good to give a reason why--in other words, the legal PAG check could perhaps return a pair <boolean, String> where the String is one reason (written out in text) why the graph is not a legal PAG if the boolean is false and null if the graph is a legal PAG. Actually, the boolean is not needed in that case.

jdramsey:
I sent out an email to the group to see if anyone else wants to comment on what I've done.

jdramsey:
This is ready for review I think

bja43:
I am not sure what to suggest in terms of coloring in the case when the graph is not a legal PAG. There are local causal discovery results that say we can still identify aspects that would color the graph. What do you think about this?

jdramsey:
I actually don't have a good sense of it. Maybe we should ask Peter?

jdramsey:
Thanks again!

PR #1461, opened 2023-01-23
Nonparanormal bug fix

jdramsey:
Fixed a bug in the nonparanormal transform where infinite transformed values (due to overflow of the inverse Normal distribution) were messing up the nonparanormal transform calculation and the histogram. Set these to the max or min of the data without the infinite values.


jdramsey:
Just want to know if this works for you now.

bja43 (approved):
Looks to be working now!

PR #1479, opened 2023-02-02
Switched from wall time to cpu time throughout code.

jdramsey:
This has been something long request, and Bryan recently pointed me to how to do it, so I went through and made the the change throughout the code.


jdramsey:
This is not a good solution to the problem, retractring.

PR #1483, opened 2023-02-04
Added seed parameter for Sem Simulation, Bayes Net Simulation, and Conditional Gaussian Simulation.

jdramsey:
If the seed is set to something other than -1, successive simulations shoudl produce the same graph and dataset.


bja43 (approved):
works

jdramsey:
Thanks!

jdramsey:
Open up the Maven tab and in Tetrad Project lifecycle, double click clean then double click test..

PR #1484, opened 2023-02-04
Changed Math. to FastMath. throughout the codebase.

jdramsey:
This is an optimization and should speed things up. Almsot all library methods use FastMath. throughout so this brings Tetrad in line with library code.


kvb2univpitt (approved):
All unit tests passed.

Issue #1478, opened 2023-02-02
Switch from wall time to CPU time throughout the code.


jdramsey:
This change was made in algcomparison, but it needs to be done everywhere for consistency.

jdramsey:
Actually a better approach would be to create (or find) a Timer class. The idea would be this:

```
Timer timer = new Timer();
timer.start();
...
timer.stop();
double cpuTme = timer.getCpuTime();
double userTime = timer.getUsertime()'
double wallTime = timer.getWallTime();
```

where the times are given in seconds with millisecond precision, ready to be printed out. What do you think @kvb2univpitt ?

jdramsey:
This change has been requested for _years_ now, should be done as soon as possible.

jdramsey:
PR: https://github.com/cmu-phil/tetrad/pull/1490

kvb2univpitt:
@jdramsey That sounds good.  Sorry for the late reply.  I must have missed the notification.

jdramsey:
Fixed, merged.

PR #1490, opened 2023-02-06
Switched to using CPU time throughout the code.

jdramsey:
Added a class, MillisecondTimes(), that returns wall time, user times, or CPU time uniformly in milliseconds. Included a method that will return whichover one is set programmatically and changed all System.currentTimeMillis() to throughout the code. The one exception is in the algcomparison Comparison class, where CPU time in particular is used.

The idea for  MillisecondTimes().timeMillis() is that the user can set which of wall time, user time, or CPU time is desired, and this method will uniformly return that time throughout the code whereever this method is used.


jdramsey:
Issue: https://github.com/cmu-phil/tetrad/issues/1478

jdramsey:
Also fixed bug in Edgewise comparison, where when comparing PAGs to PAG, edge1 ==> edge2 reports had the edges backwards.

kvb2univpitt (approved):
All unit test passed.  Looks good!

PR #1495, opened 2023-02-07
Changed the causalOrdering method in Paths (useful only for DAGs) to the validOrder method in Paths (replacing the old method and now valid for both DAGs and CPDAGs). inlined the new method throughout the code.

jdramsey:


bja43:
There appears to be a bug when running fGES and BOSS(-DC) concurrently in the algorithm comparison tool:

java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "java.util.Map.get(Object)" is null

I think we should fix this issue before completing the PR.

jdramsey:
I may have fixed this issue, PR https://github.com/cmu-phil/tetrad/pull/1500.
Could you pull from development and try again?

The reason I think I fixed it is that now your TestBoss script runs without
throwing an exception.

jdramsey:
Crap, I think I merged this somehow by accident. I don't know how. Sorry. But could you check out development and see i the issue persists? I think it's fixed now.

(How did that happen? It says I merged commit bfbd027 into development 3 hours ago, but I don't recall doing any such thing...)

Issue #1494, opened 2023-02-07
RFCI appears to give different answers on different runs on the same data. (Double check.)


jdramsey:

kvb2univpitt:
I would like to elaborate on that.  Using  IndTestProbabilistic test, RFCI should produce different graphs on multiple runs using the same data and parameters if no threshold is used.  This is because the  IndTestProbabilistic will determine if **x** and **y** are conditionally independent given **Z** based on coin flipping (random generator).  However, RFCI should produce same graphs on multiple runs using the same data and parameters with a fixed threshold.  This is because the  IndTestProbabilistic will determine if **x** and **y** are conditionally independent given **Z** based on comparing the posterior to the same threshold.

Below is a snippet to the IndTestProbabilistic code that determines the conditional independence:
```java
boolean ind;
if (threshold) {
    ind = (p >= cutoff);
} else {
    ind = RandomUtil.getInstance().nextDouble() < p;
}
```

jdramsey:
I found a source of instability--RFCI was using FasConcurrent, which doesn't guarantee that the nodes in a depth will be visited in a fixed order, because of the concurrency, and so falls prey to the issues noted with PC Stable. Solution is either to use Fas instead (without the concurrency) or to use FasConcurrent, but set isStable(true). Opting for the former. Now, running two RFCI searches off of the same SEM Simulation and comparing the two results using the Edgewise Comparison shows that the results come out consistently the same. This is with Fisher Z.

jdramsey:
Actually, 'stable' was set by default to true, which means the FasConcurrent in 'stable' mode wasn't always returning the same graph, which on the face of it should be impossible....

jdramsey:
PR: https://github.com/cmu-phil/tetrad/pull/1496

PR #1496, opened 2023-02-07
Fixing some instability in RFCI

jdramsey:
The instability was due to the use of FasDeterministic; swapped out FAS instead.

Also, cleaned up the code to use the most recent version of FciOrient. Also, removed the completete ruleset parameter, since RFCI always needs to use the complete ruleset, so this is not a parameter.


kvb2univpitt (approved):
All test passed.  Currently, RFCI does not work with Probabilistic Test.  Will fix in the future.

jdramsey:
Issue https://github.com/cmu-phil/tetrad/issues/1494

Issue #1489, opened 2023-02-06
IndTestProbabilistic: Array Index Out of Bound


kvb2univpitt:
I got ArrayIndexOutOfBoundsException when running RFCI using IndTestProbabilistic:

```java
java.lang.ArrayIndexOutOfBoundsException: 70
	at edu.pitt.dbmi.algo.bayesian.constraint.inference.BCInference.probConstraint(BCInference.java:219)
	at edu.cmu.tetrad.search.IndTestProbabilistic.probConstraint(IndTestProbabilistic.java:232)
	at edu.cmu.tetrad.search.IndTestProbabilistic.checkIndependence(IndTestProbabilistic.java:193)
	at edu.cmu.tetrad.search.IndTestProbabilistic.checkIndependence(IndTestProbabilistic.java:146)
	at edu.cmu.tetrad.search.Rfci.setMinSepSet(Rfci.java:459)
	at edu.cmu.tetrad.search.Rfci.ruleR0_RFCI(Rfci.java:299)
	at edu.cmu.tetrad.search.Rfci.search(Rfci.java:200)
	at edu.cmu.tetrad.search.Rfci.search(Rfci.java:171)
	at edu.cmu.tetrad.search.Rfci.search(Rfci.java:167)
	at edu.pitt.dbmi.pag.sampling.rfci.PagSamplingRfci$RfciSearchTask.call(PagSamplingRfci.java:189)
	at edu.pitt.dbmi.pag.sampling.rfci.PagSamplingRfci$RfciSearchTask.call(PagSamplingRfci.java:170)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)

```

jdramsey:
@kvb2univpitt This can be closed now right?

kvb2univpitt:
@jdramsey Yes.  I'll close it.  Thanks.

PR #1499, opened 2023-02-08
Fixed array out of bound error.

kvb2univpitt:
A simple fix for issue #1489.

Surprisingly, this fixed the issue where RFCI doesn't produce the same graph using a fixed threshold:

Screenshot from 2023-02-08 02-00-39


jdramsey:
Awesome! I have to try!

jdramsey (approved):
It works! I can't believe that's all i was!!!!!!!!!

I tested by making a SEM IM with two RFCI searches off of it, both into an Edgewise Comparison. In both, I set the threshold to 0.7 and selected to use a threshold. I ran it many times, no discrepancy. :-D :-D :-D

PR #1506, opened 2023-02-08
Setting 'verbose' default to 'false'.

jdramsey:
Should be false anyway but for bootstrapping this is critical.


kvb2univpitt (approved):
All test passed.  Looks like verbose is set to false by default in the GUI.

PR #1507, opened 2023-02-09
Fixing the too-verbose bootstrap output problem.

jdramsey:
1. Making sure verbose is not set to true by any algorithm. (Tests are fine.)

2. Making sure the final bootstrap output is printed out to the logs (and console output) whether verbose is set to true or not.


kvb2univpitt (approved):
Looks good!  I don't see anything print to console when verbose is off during bootstrapping.

jdramsey:
Thanks!

PR #1515, opened 2023-02-12
Making the external graph API more consistent.

jdramsey:
Renamed a bunch of methods for initial graph to external graph. Also renamed a bunch of variables similarly. Also for FGES passed the external graph in as the initial graph.

### DIFF
--- a/data-reader/pom.xml
+++ b/data-reader/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.github.cmu-phil</groupId>
         <artifactId>tetrad</artifactId>
-        <version>7.1.3-SNAPSHOT</version>
+        <version>7.2.0</version>
     </parent>
 <!--    <groupId>io.github.cmu-phil</groupId>-->
     <artifactId>data-reader</artifactId>
--- a/docs/manual/index.html
+++ b/docs/manual/index.html
@@ -20,7 +20,7 @@
 <div class="inner">
 
     <div class="header"><h1>Tetrad Manual</h1>
-        <p>Last updated: January, 18 2023</p></div>
+        <p>Last updated: 02/09/2023</p></div>
 
     <!-- Table of Contents, give each section a unique id -->
     <div
@@ -7246,7 +7246,7 @@ tabs will not appear.</p> -->
             <li>Long Description: <span id="zSRiskBound_long_desc">
                 This is the probability of getting the true model if a correct model is discovered. Could underfit.</span>
             </li>
-            <li>Default Value: <span id="zSRiskBound_default_value">0.001</span></li>
+            <li>Default Value: <span id="zSRiskBound_default_value">0.1</span></li>
             <li>Lower Bound: <span id="zSRiskBound_lower_bound">0</span></li>
             <li>Upper Bound: <span id="zSRiskBound_upper_bound">1</span></li>
             <li>Value Type: <span id="zSRiskBound_value_type">Double</span></li>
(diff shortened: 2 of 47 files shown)


kvb2univpitt:
It doesn't seem like FCI works with Fisher-Z test:

```java
java.lang.UnsupportedOperationException
	at java.util.Collections$UnmodifiableList.sort(Collections.java:1333)
	at java.util.Collections.sort(Collections.java:143)
	at edu.cmu.tetrad.search.Fas.search(Fas.java:140)
	at edu.cmu.tetrad.search.Fas.search(Fas.java:397)
	at edu.cmu.tetrad.search.Fci.search(Fci.java:198)
	at edu.cmu.tetrad.algcomparison.algorithm.oracle.pag.Fci.search(Fci.java:75)
	at edu.cmu.tetradapp.model.GeneralAlgorithmRunner.lambda$execute$2(GeneralAlgorithmRunner.java:382)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at edu.cmu.tetradapp.model.GeneralAlgorithmRunner.execute(GeneralAlgorithmRunner.java:365)
	at edu.cmu.tetradapp.editor.GeneralAlgorithmEditor$1.watch(GeneralAlgorithmEditor.java:185)
	at edu.cmu.tetradapp.util.WatchedProcess.lambda$watchProcess$0(WatchedProcess.java:159)
	at java.lang.Thread.run(Thread.java:750)
```

bja43:
What functionality would you like tested? I was able to feed fGES and PC a graph in the GUI, but I they did not seem to use the graph (their external graph variables were empty). As Kevin noted, FCI does not work if you feed it an initial graph.

jdramsey:
I guess I'd like to see if e.g. Rskew still works. I need to talk to you
guys about semantics for fges etc, what your expecting to happen. For
instance for fges I could use the external graph to set a bound graph
instead of an initial graph. That would definitely have more of an effect.
Not sure what to do about pc.. was thinking about putting that off until
the next release... But I could work on it when I get back if you guys
think it should be for this release...

cg09:
Eh?? What is an "external graph" vs an "initial graph"?   What is a "bound
graph?"

jdramsey:
Clark,

Yeah maybe you're the guy I need to talk to. We allow an external
graph for an algorithm. For a pairwise algorithm like RSkew we need
one, and the issue posted was that the user couldn't run, say, RSkew
on a given graph to orient the edges in causal-cmd. So for RSkew, the
external graph is used to determine adjacencies. For FGES we have two
options; either you can give an initial CPDAG for FES to start with
(instead of the empty graph, which should work just fine), or you can
specify a bound graph, which will restrict edges added by FGES to
adjacencies in the bound graph. I can do either, not both (I don't
think); it sounds like from what Kevin was experiencing that using the
initial graph doesn't mean much to users, since FGES is pretty good at
getting the same output no matter what graph you start FES with (see
theorems), at least for the most part, up to faithfulness that is. So
maybe the thing to do is switch that to bound graph instead of initial
graph.

For PC, I'm not sure what to do; it wasn't hooked up before, but a
natural thing to do is to take the undirected graph of the external
graph and use it in place of FAS in the algorithm, or something. I'm
not sure what exactly should be done, so I didn't do anything for this
pull request, thinking it could be addressed in a fiuture release and
a little more deliberately.

Joe

jdramsey:
@kvb2univpitt I fixed the problem in FAS that was causing the issue for you. The list of nodes being passed in was unmodifiable, so I just needed to make a modifiable copy so I could sort it.

bja43 (approved):
rSkew is taking and using input graphs correctly.

jdramsey:
Thanks!

PR #1516, opened 2023-02-13
Fixed a problem where a method was passed in an unmodifiable list of nodes and then asked to sort it. Fixed this everywhere in the same situation by making a copy of the nodes.

jdramsey:


kvb2univpitt (approved):
Yup!  This fixes the issue!

PR #1521, opened 2023-02-14
Fixed problem of SEM and Bayes simultion parameters going missing after simulation and re-opening the Simulation box.

jdramsey:


kvb2univpitt (approved):
It works!!

PR #1530, opened 2023-02-20
Final runthrough of all non-experimental algorithms.

jdramsey:
Fixed a few issues here and there--see the individual GIT commits.


bja43 (approved):
Looks good

PR #1559, opened 2023-03-21
Updated versions of BOSS and SP

jdramsey:
Bryan updated the versions of BOSS and SP and adjusted the code so that knowledge searches are done within tier (increasing the number of varaibles that can be searched).

I renamed the BOSS and SP files so that the latest ones are named BOSS and SP and the others are moved out of the way and disabled.


bja43 (approved):
changes seem to work!

PR #1568, opened 2023-03-25
some light refactoring of GST and TS

bja43:
- I refactored TS to use GSTs (and cleaned up the TS and GST code a bit).
- I think this is all working, but it definitely merits a bit of testing.
- I deleted a lot of unused code in TS which maybe @jdramsey won't like?


jdramsey (approved):
Looks great!

PR #1592, opened 2023-04-26
LiNGAM and LiNG-D code

jdramsey:
Please check the following classes:

Lingam
LingD
NRooks
TestLing

Please run the tests in TestLing to make sure they work. Still more work to do but hopefully this is a good start.


cg09:
What is NRooks?

bja43:
/**
     * Returns W, permuted rowwise by the permutation passed in through the constructor.
     * @return The row-permuted W.
     */
    public Matrix getPermutedMatrix() {
        return M.getSelection(rowPerm, colPerm);
    }

Should the comment here say rowwise and column wise?

jdramsey:
Thanks, fixed.

bja43:
I noticed that "strongestDiagonalByCols" in the LingD class is unused within the class, but is used in the Lingam class. Would it make sense to move this method to the Lingam class?

bja43:
pairsNRook and nRooks are unused (for calls to the latter, you can call pairNRooks directly)

bja43:
The class variable spineThreshold is unused (it is only used in the nRooks method call and nRooks does not utilize the passed parameter).

bja43 (approved):
Code looks good to me.

PR #1613, opened 2023-05-12
Fixed JSON graph saving/loading.

jdramsey:
Instructions, in this badly named branch, please launch the Tetrad interface, then make a random graph (a PAG would be great if you can), then save using the JSON format, then open another graph box and load as JSON format. Are they the same?


bja43:
json stores positions which do not seem to be loaded in, but everything else appears to be working properly

bja43 (approved):
saving and loading works minus the node positioning

bja43 (approved):
saving and loading works minus the node positioning

PR #1620, opened 2023-05-16
To avoid exeptions when file selection dialog closed by cancel button.

yasu-sh:
@jdramsey If you like this, adopt this proposal. When pressing cancel button, the message is displayed as LoadGraph performs. 
image


jdramsey (approved):
I approve!

PR #1626, opened 2023-05-26
IOD and some othter stuff.

jdramsey:
Implemented the IOD algorithm, plus a few other changes; I will list these in the comments to this PR.


jdramsey:
Things I did:
* I cleaned up the code for PC, CPC, FAS, and related classes, so they are squeaky clean now. You may want to look at that code. I cleaned up and expanded all of the Javadocs for those classes. The functionality changed a little, but not much. Ruben and Clark both approved the new set of parameters/defaults.
* I added a link to the Javadocs from the manual, as requested by Clark.
* I implemented the IOD algorithm at the request of Peter. The idea for this algorithm is that a new independence test was added, IndTestIod, that takes a list of independence tests as input (in the constructor), not necessarily having the same variables but having overlapping variable sets, and it does the search over these, combining their p-values where possible. The file FCI-IOD in the algcomparison package runs FCI using this new pooled independence test.
* In the Simulation box, I added a new parameter, the Probability of randomly removing a column from a dataset. If you set the Number of Runs to > 1 and this probability to > 0, you'll get a list of datasets with overlapping variables to test FCI-IOD. Then you can make a search box off of the simulation and run FCI-IOD with the Fisher Z test, and it will generate the pooled FCI graph. See attached picture for the new parameter in the Simulation box.

Screenshot 2023-05-26 at 10 35 13 AM

PR #1630, opened 2023-05-28
Some fixes and refactorings.

jdramsey:
1. Added menu items in SemImEditor and SemEstimatorEditor to let the use copy the Coefficient Matrix and Error Covariance Matrix to the clipboard.
1. Adjusted the Markov Checker to do more checks for both Markov and Faithfulness.
1. Changed all sepsets and conditioning sets throughout the codebase from lists to sets to optimize lookup.
2. Changed more list returns to set returns in EdgeListGraph and elsewhere, where it's clear a set was intended.


bja43:
Changes appear to be working as intended, but:
 - the SEM IM does not appear to match what the estimator produces using the same PM
 - faithfulness checker is misleading since you are only checking a small subset of the (in)dependencies and not all of them

jdramsey:
How would I check them all, and would that be too slow? I'll look into the estimator...it used to work...

cg09:
A query system?

jdramsey:
@cg09 What do you mean?

jdramsey:
This is for the Markov checker, so Bryan thinks we need to list all mconn(X, Y | Z) and not just all mconn(X, Y | parents(X)). I'm wondering how to make that feasible since it's exponential. I was thinking of defining mconn(X, Y | parents(X)) as "local Faithfulness" instead...

jdramsey:
I made that relabeling change.

bja43 (approved):
Changes look good.

Issue #1634, opened 2023-06-03
Data Reader Unit Tests Failed on Windows


kvb2univpitt:
The unit tests for the data reader failed when compiling on Windows OS due to the test file path.

jdramsey:
Fixed! https://github.com/cmu-phil/tetrad/pull/1635

PR #1635, opened 2023-06-03
A Fix for Unit Tests Running in Windows.

kvb2univpitt:
Fixing the unit tests failures due to not reading the correct file path in Windows.  Thanks to @yasu-sh for the idea when fixing this in causal-cmd.


kvb2univpitt:
This is a fix for issue #1634.

jdramsey (approved):
It worked! I tested on both Mac and Windows.

yasu-sh:
@kvb2univpitt Thanks for fixing tests on Windows. Now I don't have to disable test process in Maven project in IntelliJ IDEA. Good!

PR #1647, opened 2023-06-30
Work toward 7.5.0

jdramsey:


kvb2univpitt (approved):
I was able to compile and run without any issues on Java 8 and Java 17.

PR #1649, opened 2023-07-01
Nl sim mxnt dlingam

bja43:
Added Direct LiNGAM


cg09:
How does direct LingAm work with Knowledge... as with other algorithms?

bja43:
The current implementation does not allow for background knowledge, however it could be modified to account for some forms of background knowledge. Required edges might be a bit tricky, but I don't see any issue with adding functionality for tiers or forbidden edges.

jdramsey (approved):
Awesome!!! :-D

cg09:
Here is how knowledge was implemented in a github version.

jdramsey:
Here's how it was implemented in the Lingam Python package:

https://lingam.readthedocs.io/en/latest/tutorial/pk_direct.html

PR #1654, opened 2023-07-09
Adjusted Markov check and Descriptive Stats.

jdramsey:
For Markov checker, added an "all sets" option that shoudl work with #vars <= 12. If you try it with more variables than that you get a warning. Added a new Descriptive Stats menu item to the Data Editor Tools menu that puts all fo the stats into a table and allows you to copy selections from the table to be pasted into a text file or Excel. Included the lsit of constant columns and example of a 2-3 variable singular column set, if these exist, at the bottom of the tool.

Also, added a parameter, saveBootstrapGraphs, default to false. If this parameter is set to true, individual bootstrapped graphs will be saved.

### DIFF
--- a/docs/manual/index.html
+++ b/docs/manual/index.html
@@ -4466,6 +4466,28 @@ tabs will not appear.</p> -->
                 <span id="addOriginalDataset_value_type">Boolean</span></li>
         </ul>
 
+        <h3 class="parameter_description"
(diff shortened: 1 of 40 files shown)


bja43 (approved):
Everything appears to be in working order

jdramsey:
Thanks!

PR #1668, opened 2023-07-26
fixed the nl sim using svd

bja43:
the nl sim no longer throws singular matrix errors when using an rbf, its a bit quicker too


jdramsey (approved):
It works great, approve!

PR #1679, opened 2023-08-23
fixed make valid knowledge order

bja43:
Required edges should work with GRaSP and BOSS now.


jdramsey (approved):
Yep, that problem is solved, thanks!

PR #1682, opened 2023-08-24
Boss parallel

jdramsey:
Added NUM_THREADS parameter, tested required knowlege for GRaSP and BOSS and BOSS in parallel.


bja43 (approved):
thumbs up

PR #1699, opened 2023-10-13
Several fixes, updates.

jdramsey:
Several fixes--see commits. Will make a list.


jdramsey:
These fixes:

https://github.com/cmu-phil/tetrad/wiki/Forthcoming-fixes

bja43:
deleting required edges appears to be bugged, they reappear after deleting them if you check/uncheck some of the checkboxes

jdramsey:
@bja43 Can you try it now?

bja43:
The show forbidden by tiers checkbox does not appear to do anything. I put in some tiered knowledge (including forbid within tier knowledge) after which checking the box in question did not do anything. Also, I was able add an explicitly required edge that conflicted with the tiered knowledge that I specified without any sort of error being thrown.

bja43:
It also looks like you can require edges that form a two-cycle (not sure if this should be allowed since some Tetrad algorithm can indeed learn two cycles.) 

I am also seeing the same issues that I mentioned earlier for tiers, but for knowledge groups.

PR #1708, opened 2023-11-14
Changes for 7.6.1.

jdramsey:
Several recent changes.

Updates for 7.6.1 (hotfix):

1. Fixed a bug introduced in 7.6.0 where, for tiered knowledge, too many edges were forbidden per tier.
2. Fixed an issue in a snapshot build (not 7.6.0) where you couldn't make a Tiers and Edges editor if you put a knowledge box on the workbench and double-clicked it.
3. Adding an instruction in the Markov Checker editor that the contents of the table can be copied and pasted into Excel.
4. Fixed a backward compatibility issue with 7.5.0 for knowledge.
5. Stamping CPDAG algorithms with both the score used to generate them (where applicable) and standard BIC.
6. Changed the name of PC-LiNGAM to BOSS-LiNGAM.
7. Fixed problem with BOSS, GRaSP, etc. code hanging with constant columns.
8. [Code] Added code to py-tetrad to run the Markov Checker, which required moving the enum ConditioningSetType to a higher level.

### DIFF
--- a/data-reader/pom.xml
+++ b/data-reader/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.github.cmu-phil</groupId>
         <artifactId>tetrad</artifactId>
-        <version>7.6.0-SNAPSHOT</version>
+        <version>7.6.1-SNAPSHOT</version>
     </parent>
     <!--    <groupId>io.github.cmu-phil</groupId>-->
     <artifactId>data-reader</artifactId>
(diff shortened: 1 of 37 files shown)


bja43 (approved):
I checked:
- If you can load Clark’s session and run it on the first dataset, then issues 4 and 7 are fixed.
- If you can load Sisi’s dataset and run GRaSP-FCI and don’t get a slew of bidirected edges, then issue 1 is fixed.
- If you can put an empty knowledge box on the workbench and double click it and make a Tiers and Edges gadget then 2 is fixed
- If you run BOSS and it tells you the final score and BIC score, then 5 is fixed

and these things look to be working.

PR #1719, opened 2024-02-02
Added a seed parameter to BOSS and GRaSP and algorithms that use BOSS and GRASP for reproducitbility.

jdramsey:


jdramsey:
To test, simulate a dataset and add two search boxes. Choose BOSS in both. Turn off the 'use data order' parameter (which should use a random initial permutation) and select # starts = 3 for both. Set seed = 10 for both. Run BOSS. Then add a comparison box with both search boxes as parent and look at a comparison. There should be no discrepancy. Same for GRaSP, BFCI, and GRaSP-FCI.

jdramsey:
Setting seed = -1 should turn off this behavior.

jdramsey:
Here's what the parameters for BOSS look like now. Notice the new seed parameter. By default this parameter is set to -1 (no seed set) but may be set to any long value:

image

bja43 (approved):
Looks good.

yasu-sh:
@jdramsey Thanks for this feature!

jdramsey:
You're welcome!

PR #1729, opened 2024-02-12
Several updates (including some refactorings for the entire codebase).

jdramsey:
The following updates:

1. Added elapsed wall time of algorithm run to the Stats List editor if a GeneralAlgorithmRunner is a parent of it.
1. Updated the manual so that in the Search box in the interface all algorithms, tests, and scores have documentation.
1. Added a method to MarkovCheck to print out a Markov check result as a string for use in Causal Command.
1. Resolved all JavaDoc warnings.
1. Made the handling of threads more uniform, using a ForkJoinPool for everything. (This may need to be optimized.) A new ForkJoinPool is now created for each algorithm run, allowing all threads for that pool to be stopped by calling shutdownNow() on the pool.
1. Replaced special handling of the single-thread case where possible with a new ForkJoinPool with 1 thread.
1. As a result, fixed a long-standing bug in the interface where clicking the stop button would not stop all threads associated with a multi-threaded algorithm.


kvb2univpitt:
@jdramsey Bootstrapping is not working.

Do the following steps:
1. Simulate discrete data (leave default values).
2. Add a **Search** box.  Use GFIC (leave default test, score, and parameters)
3. Set bootstrap resampling to 1 and you will see a graph with no edge.  See bootstrap resampling to 100 and you'll get no graph.

jdramsey:
This is fixed. I realized using a singleton ForkJoin to store a ForkJoinPool for all algorithm runs was a mistake; ForkJoinPools should be created for each algorithm so that threads specific to that algorithm run can be stopped using pool.shutdownNow(). For the ForkJoin class was deleted and all handling of concurrency returned to individual algorithms. This is the revised list of issue fixes:

1. Added elapsed wall time of algorithm run to the Stats List editor if a GeneralAlgorithmRunner is a parent of it.
2. Updated the manual so that in the Search box in the interface all algorithms, tests, and scores have documentation.
3. Added a method to MarkovCheck to print out a Markov check result as a string for use in Causal Command.
4. Resolved all JavaDoc warnings.
5. Made the handling of threads more uniform, using a ForkJoinPool for everything. A new ForkJoinPool is created for each run of each algorithm, allowing all threads for that pool to be stopped by calling shutdownNow() on the pool when the current thread is interrupted.
6. For parallelized algorithms, add (where not already available) a parameter to specify the number of threads to use.
7. For bootstrapping, also added a parameter for the number of threads to use (for the bootstrapping itself).
8. This fixes a long-standing bug in the interface where clicking the stop button would not stop all threads associated with a multi-threaded algorithm.

jdramsey:
I cleaned up the bootstrapping code a bit and commented it.

kvb2univpitt (approved):
Looks good!  Looks like bootstrapping is working again.

PR #1732, opened 2024-02-20
Some code adjustments.

jdramsey:
Moved the session package to gui. Did a number of other things. See commit messages.


jdramsey:
Note--Peter had wanted me to fix logging in PC, which I just did in the last couple of commits. I haven't gone through the entire interface though to make sure it's fixed everywhere, which I should.

jdramsey:
I made sure the Markov checker simulation doesn't print verbose output, so that covers a lot of algorithms with the logging.

kvb2univpitt (approved):
It seems to work!

PR #1734, opened 2024-02-26
Mainly documentation.

jdramsey:
Did a lot of documenation updates in tetrad.search.


bja43:
The GUI seems to freeze up when editing equations in the generalized SEM PM. My client froze in the process of changing the errors to Uniform(-1,1) when editing a generalized SEM for a Y structure. If the client doesn't immediately freeze, try clicking done, reopen the PM, and edit an equation.

jdramsey:
This could be a race condition. I took out all of the synchronized keywords from that code (since they weren't needed). Maybe that solves the problem.

kvb2univpitt (approved):
Just review this with Joe.  Seems to be working as expected.

PR #1737, opened 2024-02-29
Added IDA minimum beta values to the MarkovCheck and MarkovCheckEditor classes.

jdramsey:
To test this, in the Tetrad interface, run a CPDAG algrothm like BOSS and then add a comparison box and do a Markov check. There should now (int his branch) be a new column in the implied independencies and implied dependencies tabs with these min betas. The IDA beta for x to y is listed in the row for x _||_ y | parents(x) (or x ~_||_ y | parents(x). For the independence case these numbers should be small, but for the dependence case there should be a number of large betas.


cg09:
What is an IDA Beta.

Issue #1740, opened 2024-03-12
Cancelling Search Task Should Not Show Empty Graph


kvb2univpitt:
Cancelling a search task should not switch to a Graph Panel showing an empty graph.

kvb2univpitt:
Fixed.

PR #1743, opened 2024-03-12
Added a field to indicate if a stop button has clicked.

kvb2univpitt:
A fix for issue #1740.  Clicking on the cancelling button during search will stop the search and stay on the Parameter panel instead of switching to a Graph panel with an empty graph.


jdramsey (approved):
Looks good; I can't trick it up. I tried various algorithms with and without bootstrapping.

Issue #1741, opened 2024-03-12
Progress Button Should Go Away When Error Message Pops Up


kvb2univpitt:
An error message alert will appear when there's something wrong while running search algorithm.  The **"Processing (Click to Stop)"** button should be hidden when the error message appears.

jdramsey:
This is done.

PR #1744, opened 2024-03-12
Propagate exception from running tasks to the process watcher.

kvb2univpitt:
Propagate exception from running tasks to the process watcher. so that exception could be catched.  The changes also includes a fix for issue #1741.


jdramsey (approved):
Oh this fixes the MGM bootstrap problems! Hurray!

Also, I created a dataset and added a copy of one of the columns to the dataset and ran FASK with and without bootstrapping. In each case, the singularity exception was reported, yay!

This is working, thanks!

Issue #1745, opened 2024-03-13
Bug reports from Peter...


jdramsey:
The following are bugs in tetrad 7.6.3 (attached session). 

(1) The model statistics are wrong. It has a p-value of 1 even though the degrees of freedom are 7 and the chi-squared is not near 0. It also says the same p-value for multiple data sets generated by the DAG.

(2) The X3 -> X5 edge is not dotted, but I am sure that while X3 is an ancestor of X5, it does not eliminate the possibility of latents. (Previous versions of tetrad do not make the X3 -> X5 edge green.

(3) The help for interpreting edges in FCI is also available in PC, where it is misleading because it says that solid edges mean that there is no unmeasured confounder. In addition, it should not say "More information on edge types and colorings", since the colorings are no longer used.

jdramsey:
Responses:

(1) This has been fixed in the development branch but has not yet been published.

(2) This is actually working, though the PAG coloring does need to be applied intentionally to see it. I will think about whether the PAG coloring can be automatically applied to latent variable algorithms instead...

(3) Well, in the search box, I parsed the annotation to determine if an algorithm is an LV algorithm and displayed the button for the edge type explanation only if it is. I'll make a PR for this. Also, this information is not available in the Graph box, so the button was removed there. The text will be adjusted for this PR.

jdramsey:
Well, for (2), it was not hard to add the PAG coloring back in automatically for LV algorithms in the search box, and I adjusted the PAG coloring menu item for this.

It is still unclear what the behavior should be for the Graph boxes. To be clear, the issue is that the PAG coloring and LV results can be shown in these graph boxes. Still, information about which algorithm generated the graphs is not available, so the annotation for the algorithm cannot be checked. Will ask around.

PR #1746, opened 2024-03-13
Responding to bug reports from Peter.

jdramsey:
https://github.com/cmu-phil/tetrad/issues/1745

Added a check of the annotations of an algorithm to determine if the algorithm has been declared a LV algorithm.

Used this check to show PAG coloring in the search box for LV algorithms like FCI and for no other algorithms, since it doesn't apply to other graph types.

Fixed the PAG coloring menu item so it works with this update.

Took away the FCI edge type info button from other graph boxes, as the type of algorithm that generated the graph is not available there.


kvb2univpitt (approved):
No issue with the changes.

PR #1748, opened 2024-03-19
User window preferred size and location.

jdramsey:
When the user resizes or moves the screen, the preferred size and location should be updated and used next time the app is started.


kvb2univpitt (approved):
Works great!

PR #1749, opened 2024-03-20
Updated/corrected IDA and added a UI for it in the Swing Tetrad interface.

jdramsey:
I will attach a sample session that can be used to test this in this branch.


bja43:
the min/max diff stats should have "average" in the name, other than that, looks good.

bja43 (approved):
see comment in discussion

PR #1754, opened 2024-04-05
Algomparison UI

jdramsey:
This has an implementation of a model and editor for the Algcomparison API, in the Tetrad Java Swing interface.

You can get to it (in this branch) by placing a Compare box on the workbench and double clicking it. It takes no parents currently.

Lots of text is generated; hopefully this is adequate. The one thing we don't have anywhere are descriptions of the simulations, so those aren't included. We'd have to write those and put them somewhere.

Let me know if it's OK.


bja43:
I am not pulling in case Kevin want to review first.

PR #1755, opened 2024-04-05
Add public getPValue method for IndTestChiSquare

vbcwonderland:
Introduce `public Double getPValue(Node x, Node y, Set<Node> z)` 

Because `getPValue`is not within `IndependenceTest` public interface. 
Currently, for ChiSquare, getPVal() has a different function signature in `ChiSquareTest` class 's `ChiSquareTest.Result.` Usage example:

(in `IndTestChiSquare.java`)

```
...
private final Map<IndependenceFact, ChiSquareTest.Result> facts = new ConcurrentHashMap<>();
...
ChiSquareTest.Result result = this.facts.get(new IndependenceFact(x, y, _z));
            return new IndependenceResult(new IndependenceFact(x, y, _z), result.isIndep(), result.getPValue(),
                    getAlpha() - result.getPValue());
```

By adding this new method `getPValue(Node x, Node y, Set<Node> z)`  in `IndTestChiSquare`, we can access its `private final Map<IndependenceFact, ChiSquareTest.Result> facts = new ConcurrentHashMap<>();`  and get corresponding pvalue in need.

### DIFF
--- a/tetrad-lib/src/main/java/edu/cmu/tetrad/search/test/IndTestChiSquare.java
+++ b/tetrad-lib/src/main/java/edu/cmu/tetrad/search/test/IndTestChiSquare.java
@@ -246,6 +246,21 @@ public final class IndTestChiSquare implements IndependenceTest, RowsSettable {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jdramsey (approved):
Pefect!

PR #1756, opened 2024-04-05
add IndTestChiSquare case

vbcwonderland:
Since we introduced [#1755 ](https://github.com/cmu-phil/tetrad/pull/1755)
In `getLocalPValues` now we can also consider the case for `IndTestChiSquare`


jdramsey (approved):
Looks good!

PR #1760, opened 2024-04-18
Null checks in EdgeListGraph

vbcwonderland:
Null guards to avoid NPE for `getAdjacentNodes` and  `addEdge`.


vbcwonderland:
Screenshot 2024-04-17 at 9 04 39 PM

jdramsey (approved):
I can see why you would need that. Looks good.

PR #1769, opened 2024-05-14
Finished up experimental LV-Lite and cleaned up the rest of the LV code.

jdramsey:


bja43 (approved):
There are too many options for the grid search to exhaustively check them all, but I was able to get a few simple comparisons up and running without issue.

PR #1773, opened 2024-05-21
Update getLocalIndependenceFacts to check on independence

vbcwonderland:
getLocalIndependenceFacts should check on independence rather than validity.


jdramsey (approved):
Good.

PR #1774, opened 2024-05-21
Introduce shuffle trick to get more local P-values

vbcwonderland:
Sampling by shuffle threshold and get P Values for each shuffle group. The result `pVals_list` will then being used in other methods. 
For example: generate each of its own coresponding Anderson-Darling Test P Value.

### DIFF
--- a/tetrad-lib/src/main/java/edu/cmu/tetrad/search/MarkovCheck.java
+++ b/tetrad-lib/src/main/java/edu/cmu/tetrad/search/MarkovCheck.java
@@ -261,6 +261,39 @@ public class MarkovCheck {
     }
 
     /**
+     * Get Local P values with shuffle threshold provided.
+     * @param independenceTest
+     * @param facts
+     * @param shuffleThreshold
+     * @return
+     */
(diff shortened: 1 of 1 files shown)


jdramsey (commented):
Should this: int shuffleTimes = (int) Math.ceil(1 / shuffleThreshold);
be this?: int shuffleTimes = (int) Math.ceil(sampleSize / shuffleThreshold);

for some sample size?

vbcwonderland:
Hi Joe, 

I think it should be `(int) Math.ceil(1 / shuffleThreshold)` because the `shuffleThreshold` here is a decimal, e.g. 0.3. 
Then in order to cover all samples at least statistically for once (i.e. no shuffle, use the whole sample), we should do `int) Math.ceil(1 / shuffleThreshold)` so that it can achieve the same.  e.g. 1 / 0.3 = 3.333 -> round up to 4 times, would make sure if we take 30% of our current data for 4 times, we would expect to use all data points for once. 

Best,
Verity

jdramsey:
Oh i see what you mean... let me go over this again now.

jdramsey (approved):
Looks good.

PR #1787, opened 2024-06-25
Resolve NaN ADTest P value from empty independent facts cases

vbcwonderland:
We were having NaN as ADTest P Vals for cases when the independence facts list is empty. This would not show up in large graph but would occur in small ones. 
Now with the fix, this would no longer be an issue.


jdramsey (approved):
Oh I see... OK. This is fine.

PR #1790, opened 2024-06-26
Fix node replacement issue for estimated graph for both local and global graph

vbcwonderland:
Local graph is doing node replacement in local graph method level. This is also how we do it for local graph confusion. 
Global graph node replacement should be done in whole graph level. This is also how we do it for Adj confusion and Arrowhead confusion.


jdramsey (approved):
Good.

PR #1793, opened 2024-07-09
Update LocalGraph Confusion matrix calculation

vbcwonderland:
In the check orientation step, instead of updating FN when Estimated graph is `--`, we decide to do nothing rather than undapting. This is because `--` means unknown.


jdramsey (approved):
Yes--you should explain this to Peter as well.

PR #1799, opened 2024-07-16
Include specify file for test check nodewise markov

vbcwonderland:
Instead of randomly generated true graph each time, now enable specify provide certain file for the test.
Also added one file example `testTrueGraphForCheckNodewiseMarkov.txt` for reference of file format.


jdramsey (approved):
Approving--but with a question:

Does Peter want to also supply his own dataset?

vbcwonderland:
Hi Joe, 

Thanks for the review and merge! 

I think Peter wants to use this functionality so he can do multiple tries. This is why I provided a sample .txt file for his reference. 

Best,
Verity

PR #1826, opened 2024-09-24
joe-work-2024-9-20c

jdramsey:


kvb2univpitt (approved):
Looks good.  I was able to compile and run on JDK 21.

PR #1857, opened 2025-03-24
WIP: testGaussianDAGPrecisionRecallForLatentVariableOnLocalOrderedMarkov

vbcwonderland:
simulation test for latent variable on LocalOrderedMarkov serving the LV-Light paper. 

Besides regular tests on `ap`, `ar`, `ahp`, `ahr`, `lgp`, `lgr`,
we also included stats for 
`andersonDarlingA2` (Anderson Darling Test A-Square),
`kSPvalue` (Kolmorogov-Smirnov p-value), 
`fractionDep` (fraction of dependent judgments), 
 and `numTests` (number of tests generateResults actually did). 
These stats would be important for our future report generation.

### DIFF
--- a/tetrad-lib/src/test/java/edu/cmu/tetrad/test/TestCheckNodewiseMarkov.java
+++ b/tetrad-lib/src/test/java/edu/cmu/tetrad/test/TestCheckNodewiseMarkov.java
@@ -157,6 +157,61 @@ public class TestCheckNodewiseMarkov {
         System.out.println("Rejects size: " + rejects.size());
     }
 
+    public static void testGaussianDAGPrecisionRecallForLatentVariableOnLocalOrderedMarkov(File txtFile, double threshold, double shuffleThreshold, double lowRecallBound) {
+        Graph trueGraph = GraphSaveLoadUtils.loadGraphTxt(txtFile);
+        System.out.println("Test True Graph: " + trueGraph);
+        System.out.println("Test True Graph size: " + trueGraph.getNodes().size());
+
+        SemPm pm = new SemPm(trueGraph);
+        // Parameters without additional setting default tobe Gaussian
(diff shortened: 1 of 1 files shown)


jdramsey (commented):
Actually, you can call this method, just generateResults(true) with one argument:

    /**
     * Generates all results, for both the Markov and dependency checks, for each node in the graph given the parents of
     * that node. These results are stored in the resultsIndep and resultsDep lists. This should be called before any of
     * the result methods. Note that only results for X _||_ Y | Z1,...,Zn are generated, where X and Y are in the
     * independenceNodes list and Z1,...,Zn are in the conditioningNodes list.
     *
     * @param clear True, if the results should be cleared before generating new results; otherwise, the new results are
     *              appended to the existing results.
     * @see #getResults(boolean)
     */
    public void generateResults(boolean clear) {
        generateResults(true, true);
        generateResults(false, false);
    }

vbcwonderland:
> Actually, you can call this method, just generateResults(true) with one argument:
> 
> ```
> /**
>  * Generates all results, for both the Markov and dependency checks, for each node in the graph given the parents of
>  * that node. These results are stored in the resultsIndep and resultsDep lists. This should be called before any of
>  * the result methods. Note that only results for X _||_ Y | Z1,...,Zn are generated, where X and Y are in the
>  * independenceNodes list and Z1,...,Zn are in the conditioningNodes list.
>  *
>  * @param clear True, if the results should be cleared before generating new results; otherwise, the new results are
>  *              appended to the existing results.
>  * @see #getResults(boolean)
>  */
> public void generateResults(boolean clear) {
>     generateResults(true, true);
>     generateResults(false, false);
> }
> ```

thank you Joe, included this update in the new commit. :D

jdramsey (approved):
This is fine.

PR #1860, opened 2025-04-04
Markovcheck: Convert true graph to PAG before comparison

vbcwonderland:
convert true graph to PAG and then compare with the estimated PAG


jdramsey (approved):
Looks good.

PR #1862, opened 2025-04-08
Introducing multiple FCI simulations BossFci, GaspFCI, FCI, FCIMax, RFCI

vbcwonderland:
WORK IN PROGRESS, no need for review yet. 
Although FCIMax already set as PAG, it still complains about input is not a legal MAG. Will continue fix it tmr

currently it looks like this:

Screenshot 2025-04-08 at 2 33 11 AM


jdramsey:
Oh thanks--

PR #1864, opened 2025-04-09
Remove and ignore .DS_Store

vbcwonderland:
This PR removes a tracked `.DS_Store` file from the repository and adds it to the `.gitignore` to prevent it from being committed in the future.

`.DS_Store` is a system file automatically generated by macOS to store folder metadata. It is not relevant to the project and should not be version-controlled.

This change is unrelated to any current feature work and is intended to keep the repository clean and platform-agnostic.


jdramsey (approved):
Good idea.

PR #1870, opened 2025-04-26
Add new simulators, search methods, and scoring options

jdramsey:
### Description

This pull request introduces several major updates and features:
- Added new simulators: functional-causal, additive-noise, post-non-linear, and causal-perceptron.
- Implemented Dm/DmFcit latent-discovery search method.
- Introduced new CCA-based scoring options, such as CCA-BIC and BF-LRT.
- Deprecated the MGM algorithm, which has been relocated to the legacy section.

These updates enhance functionality and address contemporary modeling needs.

### DIFF
--- a/tetrad-lib/src/main/resources/docs/manual/index.html
+++ b/tetrad-lib/src/main/resources/docs/manual/index.html
@@ -360,6 +360,17 @@
     <p>Isolates the subgraph of the input graph involving all and only
         latent variables.</p>
 
+    <h3 id="ui-indicator-expansion">Discrete-to-Indicator Expansion</h3>
+    <p>
+        In <b>Data ▸ Transform</b>, the new <b><i>Indicator</i></b> button converts
+        a discrete column into <i>k−1</i> {0,1} dummies preserving rank.
+    </p>
+
+    <h3 id="ui-bootstrap-options">Bootstrap “Sample w/out Replacement”</h3>
+    <p>
+        The bootstrap dialog now offers <b>Without replacement</b>.
+        Useful for exact subsampling when the sample size is small.
+    </p>
 
     <h2>Other Graph Box Functions</h2>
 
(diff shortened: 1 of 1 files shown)


cg09:
Why send MGM away?

What is CCA?  Are new algorithms documented in the manual?

jdramsey:
It became incompatible with the codebase and I couldn't for the life of me
figure out how to update it. The original coders are no longer around.

jdramsey:
(I did try. I spent a couple of days on it.)

jdramsey:
By the way, it's in a previous version; if anyone wants to use it, they can
download that version and run it. I documented that.

Joe

cg09:
Never mind the old cold. Input is data and a set of one factor
models. Estimate the values of each latent for each case (Its a standard
problem, I think). Then run a search over the values of the latents.

jdramsey:
It's actually not a standard problem. The place where I got stuck was where a complicated Matlab function was clumsily translated into Java. That's what I couldn't figure out how update; the matrix algebra there was fragile.

jdramsey:
Yeah that's what the issue was. We updated out matrix library because the old one had fallen out of repair and was not longer supported. The Matlab translation of that search could not be translated into the new matrix algebra class. Everythign else in Tetrad could be. In any case, if someone wants to use it they can, just not in the most recent version.

cg09:
Here is Chat gpt code for the problem using principal components:
import org.apache.commons.math3.linear.*;
import java.util.Arrays;

public class PCAFactorApproximation {
    public static void main(String[] args) {
        // Example data: rows = cases, columns = observed variables
        double[][] data = {
            {2.5, 2.4},
            {0.5, 0.7},
            {2.2, 2.9},
            {1.9, 2.2},
            {3.1, 3.0},
            {2.3, 2.7},
            {2, 1.6},
            {1, 1.1},
            {1.5, 1.6},
            {1.1, 0.9}
        };

        RealMatrix X = MatrixUtils.createRealMatrix(data);
        RealMatrix XCentered = centerColumns(X);

        // Compute covariance matrix
        RealMatrix covariance = new
Covariance(XCentered).getCovarianceMatrix();

        // Eigen decomposition
        EigenDecomposition eig = new EigenDecomposition(covariance);
        RealMatrix eigenvectors = eig.getV();
        RealVector firstPC = eigenvectors.getColumnVector(0);

        // Project data onto first principal component
        double[] latentScores = XCentered.operate(firstPC).toArray();

        // Output estimated latent variable (first principal component)
        System.out.println("Estimated latent variable scores:");
        Arrays.stream(latentScores).forEach(score ->
System.out.printf("%.4f%n", score));
    }

    // Helper: center the columns (zero mean)
    private static RealMatrix centerColumns(RealMatrix matrix) {
        int rows = matrix.getRowDimension();
        int cols = matrix.getColumnDimension();
        double[][] centered = new double[rows][cols];

        for (int j = 0; j < cols; j++) {
            double mean =
Arrays.stream(matrix.getColumn(j)).average().orElse(0);
            for (int i = 0; i < rows; i++) {
                centered[i][j] = matrix.getEntry(i, j) - mean;
            }
        }
        return MatrixUtils.createRealMatrix(centered);
    }
}

jdramsey:
Principle components was the issue. There was an optimization of some sort. I couldn't follow the code. I can dig up the code and send it to you if you want to take a shot at it.

PR #1876, opened 2025-05-17
Update SemBicScore.java

bja43:
When passing a Cov Mat on the Fly to the SEM BIC score, it will compute the full matrix which seems undesirable.

This should be a fix to that situation.

(as far as I can tell)
The only time this.matrix is accessed is when this.dataModel is null but this.dataModel is never null when a Cov Mat on the Fly is passed. Accordingly my fix is to only calculate this.matrix when this.dataModel is null.


jdramsey (approved):
Good.

PR #1883, opened 2025-06-01
Check Nodewise Markov add record for true graph's PAG

vbcwonderland:
This PR adds record for PAG version of the true graph. 
Also fixed the issue that for true graph DAG to PAG should be calculated once per simulation. 
Screenshot 2025-06-01 at 12 43 21 PM


jdramsey (approved):
Looks good!

PR #1913, opened 2025-06-27
Make allIndependenceFacts a field in MarkovCheck

vbcwonderland:
This PR introduces `allIndependenceFacts` as a new field in the `MarkovCheck` class.

Previously, `allIndependenceFacts` was a local variable within the `generateResults `method. By promoting it to a class-level field, we enable its reuse in downstream methods. In particular, this change supports optimization efforts in the Nodewise MarkovCheck workflow.


jdramsey (approved):
Looks good.

PR #1915, opened 2025-06-29
optimise msep check for target node in  checkIndependenceForTargetNode

vbcwonderland:
use new field variable `allIndependenceFacts`  in `checkIndependenceForTargetNode`. This way, the `allIndependenceFacts` would be depending on the generated results of Markov Checker depending on the conditioning set type.


jdramsey (approved):
This looks good.

PR #1917, opened 2025-07-03
checkIndependenceForTargetNode care about odering for OLMM case

vbcwonderland:
`checkIndependenceForTargetNode` when getting back a list of `IndependenceFact` for target node, needs to consider sequence of x, y position when the Markov Checker is of `ORDERED_LOCAL_MARKOV` condition set type.


jdramsey (approved):
That's correct.

PR #1922, opened 2025-07-16
include whole graph precession and recall check for adjacency, arrowhead, tail, and circle

vbcwonderland:
Add `getPrecisionAndRecallWholeGraph` that works on true graph and estimated graph for global precissions and recalls.


jdramsey (approved):
Looks good.

jdramsey (commented):
Trying again...

